PR #9, opened 2025-01-22
Allows to use the board in another module

DivadNojnarg:
@nbenn
This PR ensures we can use the board from another parent module. To insert or remove nodes, specifying id alone does not allows to capture a potential parent module. We can't also do ns(id) as this would duplicated the lower level id. A solution is to pass `ns(NULL)` which just returns `<parent_id>-<current-id>`.

Other changes (spaces, indent) are due to my R code formatter.

You can test from blockr.ui@board-v2 branch running the `inst/examples/demo/app.R` app.


nbenn (approved):
Lgtm, thanks!

PR #22, opened 2025-01-30
Fix 18 improve UI

DivadNojnarg:
- Fix flex class issues for the toolbar container.
- Added stick-top capability.
- Tweak button icons and text.
- Margin and padding utils.
- 
Note: some code got formatted by my IDE formatter (air).


nbenn (approved):
Many thanks! We have a bit of a problem with your text editor being unhappy with my coding style and creating noisy diffs. But that's a problem for later I guess. I'll merge.

PR #28, opened 2025-02-04
slow text input

DivadNojnarg:
@nbenn Feel free to test and let me know. The input contains its own JS dependency (I did not want to make the dependency part of the blockr.core UI as people may customize it and could lose the dependency by doing so). Also create a R wapper for it where you can pass the `debounce` rate. This is forwarded to JS in the `getRatePolicy` method.


nbenn (approved):
Many thanks!

PR #34, opened 2025-02-06
Block state update policy

### LINKED ISSUE #33: For block state variables, determine correct update policy
This is the same issue as cynkra/blockr.dplyr#13: for JSON "restoration" we want `ignoreNULL = TRUE` and to satisfy "user expectations" when manipulating such values we want `ignoreNULL = FALSE`. Maybe the correct update policy is `ignoreNULL = FALSE, ignoreInit = TRUE`?

cc @DivadNojnarg.

nbenn:
Intended to fix #33.


nbenn:
@DivadNojnarg I believe this "update policy" works correctly both for restoring and clearing the selection. Let me know if you see any remaining issues.

Issue #32, opened 2025-02-05
Lock down plugin API


nbenn:
Pass read-only copy to plugins/callbacks, maybe something like

```r
x <- reactiveValues(a = 1)
y <- `class<-`(`[[<-`(unclass(x), "readonly", TRUE), class(x))

isolate({y$a <- 2})
#> Error in `$<-`:
#>  ! Can't modify read-only reactive value 'a'
isolate({x$a <- 2})
isolate(x$a)
#> [1] 2
isolate(y$a)
#> [1] 2
```

PR #35, opened 2025-02-06
Read-only rv in board server

nbenn:
Plugins should have read-only access to `rv`.


nbenn:
@DivadNojnarg this is ready for you to check out. Mostly what we discussed yesterday.

- `rv` passed to plugins/callbacks is "read-only"
- you can pass any (reactive) object(s) from the parent scope and they will be available in plugin modules/callbacks (freely modifiable)
- the board server return value changed slightly to include (modified) objects that were passed from the parent scope
- `rv` is returned (as read-only) to the parent scope unnamed, in position 1

https://github.com/cynkra/blockr.core/blob/dcaea42b88e8f051578b7782d8c56313a3db68f3/R/board-server.R#L166-L169

The board server return value might need some refinement. But this is what we have right now.

Happy to take suggestions/feedback & let me know if you run into any issues.

All in all a low-priority change. I just wanted to get it you there for you to try out.

nbenn:
Fixes #32.

PR #39, opened 2025-02-20
Update docs

MikeJohnPage:
This branch is a work in progress and is not ready for review. I'm opening the PR to initiate discussion on the docs.


MikeJohnPage:
@nbenn I see a `gh-pages` branch has been created, so presumably the plan is to use a pkgdown site. For this, do you have a preference between vignettes or pkgdown articles for the docs?

nbenn:
@MikeJohnPage A Gh action has been building the pkgdown site for some time now. Just added a [link](https://cynkra.github.io/blockr.core/) to the repo metadata.

> do you have a preference between vignettes or pkgdown articles for the docs?

At this point I'm open to anything. Right now, we just have to start creating some content I feel. We can look into fine-tuning "presentation" later. In that sense, if you feel like starting to work on any variant of long-form documentation, have at it!

An obvious vignette/article/... would be something like "How to create a block". Happy to discuss if you have other ideas.

MikeJohnPage:
Another vignette I would like to write, or should be written is testing your blocks. I'm just playing with `shiny::testServer()` now, and it is very powerful, but some things aren't obvious, such as that you need set the `app = my_block_constructor()$expr_server`

nbenn:
Yeah, that'd be helpful for block authors for sure!

One thing that is important to understand: a block consists of 2 modules: the "block" module which handles functionality that is shared across blocks and a nested "expr" module that essentially deals with building the block expression (using user inputs etc.).

The block dev only provides the expr module, and if you want to test only that, you can do so, by passing only the block "expr" server function to `shiny::testServer()`.

(The `blockr.core::block_expr_server()` helper is a simple "getter" function:

```r
function(x) {
  x[["expr_server"]]
}
```

the just retrieves whatever you passed as `server` in your block ctor.)

If you want to test your "expr" module in the context of the entire "block" module, you can also do so. For that they need to pass the correct `block_server()` implementation to `shiny::testServer()` which involves S3 dispatch. For 'blockr.core' tests I have a small (hacky) utility that does this:

```r
get_s3_method <- function(generic, object) {

  for (cls in class(object)) {
    res <- utils::getS3method(generic, cls, optional = TRUE)
    if (is.function(res)) {
      return(res)
    }
  }

  stop("No function found for generic `", generic, "()` and classes ",
       paste_enum(class(object)))
}
```

Maybe this could be exported as well for use in extension package unit tests. I'm sure there are edge-cases, where this will return a wrong function, but for the well-behaved cases as we're dealing with it's probably fine.

DivadNojnarg:
@nbenn I think we're good to go live. Let me know if you find anything wrong. @MikeJohnPage As I am off next week, feel free to open new small PRs to complement and tweak these vignettes.

MikeJohnPage:
I've already made a few tweaks and added a new testing vignette, so for now, I'm happy to go live. If I can find some spare time I will do another scan through your latest changes @DivadNojnarg. Thanks!

PR #42, opened 2025-02-24
Refactor board update mechanism

### LINKED ISSUE #41: Reduce communication between plugins
In order to require less communication between plugins, we can initiate any board change (add/rm blocks/stacks/links) from any plugin via an `update` `reactiveVal()` module argument.

nbenn:
Fixes #41.


nbenn:
@DivadNojnarg I added a `plugins` class with the intention of making things like UI generation a bit more seamless. You can now do things like

https://github.com/cynkra/blockr.core/blob/c02af6e6c0f93e5786ec17f53c4560d876830cef/R/board-ui.R#L21-L32

I'm not sure I succeeded (i.e. the complexity/usefulness tradeoff is not super). Try it out and let me know what you think.

PR #43, opened 2025-02-25
UX enhancements II

nbenn:
Continues #40.


nbenn:
@DivadNojnarg There's one known issue (which got me during the demo yesterday): when you remove a block from the block-specific module, the lack of proper cleanup leads to reactives/observers being triggered in the module that corresponds to the removed block and this leads to errors.

Reproducible example:

```r
serve(
  new_board(new_dataset_block("BOD"))
)
```

and then hit the "remove block" action.

I'm looking into cleaning up things like inputs/outputs when a module is removed. Not sure hot to tackle all observers that are created in a module. I'd have to register all of them in a centralized "directory" to be able to clean them up. Seems a bit messy, so I'll start with inputs/outputs.

PR #54, opened 2025-03-28
Will fix format.block: can't print block that is not part of a packag…

### LINKED ISSUE #53: format.block: can't print block that is not part of a package
@nbenn 

I am defining a dummy block for a vignette documentation and don't want that block to appear in the package `R` folder.

```r
new_dummy_block <- function(text = "Hello World", ...) {
  new_data_block(
    function(id) {
      moduleServer(id, function(input, output, session) {
        list(
          expr = reactive(quote(text)),
          state = list(text = text)
        )
      })
    },
    function(id) {
      tagList()
    },
    class = "dummy_block",
    ...
  )
}
```

I can't print it:

```r
new_dummy_block()
Error in `paste0()`:
! cannot coerce type 'closure' to vector of type 'character'
```

It seems that we can't print blocks that are not part of a package. 

Putting a breakpoint inside `format.block`, we probably just need an `if` statement when `ctor_pkg` is NULL.

DivadNojnarg:
Fixes #53

### DIFF
--- a/R/block-class.R
+++ b/R/block-class.R
@@ -520,7 +520,7 @@ format.block <- function(x, ...) {
     out <- c(out, "Stateless block")
   }
 
-  if (!is.null(attr(x, "ctor_pkg"))){
+  if (!is.null(attr(x, "ctor_pkg"))) {
     ctor <- paste0(attr(x, "ctor_pkg"), "::", attr(x, "ctor"), "()")
   } else {
     ctor <- paste(deparse(attr(x, "ctor")), collapse = "\n")
--- a/R/block-class.R
+++ b/R/block-class.R
@@ -523,7 +523,7 @@ format.block <- function(x, ...) {
   if (!is.null(attr(x, "ctor_pkg"))) {
     ctor <- paste0(attr(x, "ctor_pkg"), "::", attr(x, "ctor"), "()")
   } else {
-    ctor <- paste(deparse(attr(x, "ctor")), collapse = "\n")
+    ctor <- "<local function>"
   }
   c(out, paste("Constructor:", ctor))
 }
--- a/R/block-class.R
+++ b/R/block-class.R
@@ -520,10 +520,12 @@ format.block <- function(x, ...) {
     out <- c(out, "Stateless block")
   }
 
-  c(
-    out,
-    paste0("Constructor: ", attr(x, "ctor_pkg"), "::", attr(x, "ctor"), "()")
-  )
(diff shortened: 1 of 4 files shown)


nbenn (commented):
Thanks for the PR. On first sight, I think deparse-printing the entire ctor is a bit much (not-so relevant) info for printing. With a simple example like this it's not too bad, but more complex functions might be a bit overwhelming. What do you think? We could also just print some constant like `<local function>` or so.

R/block-class.R
```diff
@@ -523,7 +523,7 @@ format.block <- function(x, ...) {
   if (!is.null(attr(x, "ctor_pkg"))) {
     ctor <- paste0(attr(x, "ctor_pkg"), "::", attr(x, "ctor"), "()")
   } else {
-    ctor <- paste(deparse(attr(x, "ctor")), collapse = "\n")
+    ctor <- "<local function>"
```
DivadNojnarg, line 526: @nbenn yes that's probably enough

DivadNojnarg:
>(Just a side remark: with such non-pkg ctors, other issues might pop us as well. I'm not sure what happens if you try to serialize such a block for example.)

My initial idea was to try to register a local block within the vignette code and show that the registry is updated. That's where I found this issue, but it seems that despite fixing this one, the registration does not work.

nbenn:
@DivadNojnarg maybe that resolves your issue. Let me know if not and attack a small example. Happy to try fixing further.

DivadNojnarg:
That's my example:

```r
new_dummy_block <- function(text = "Hello World", ...) {
  new_data_block(
    function(id) {
      moduleServer(id, function(input, output, session) {
        list(
          expr = reactive(quote(text)),
          state = list(text = text)
        )
      })
    },
    function(id) {
      tagList()
    },
    class = "dummy_block",
    ...
  )
}

register_dummy_blocks <- function() {
  register_blocks(
    c("new_dummy_block"),
    name = c("dummy block"),
    description = c("A block that does nothing"),
    package = NULL,
    overwrite = TRUE
  )
}

register_dummy_blocks()
```

It does not seem to work with the latest commit. Also, independent from this commit, it errors when I don't pass `package = NULL` (! is_string(package) is not TRUE).

nbenn:
@DivadNojnarg if you pass a "local" function as block constructor it should be passed not as string, but as function.

```r
register_blocks(
  c(new_dummy_block),
  name = c("dummy block"),
  description = c("A block that does nothing"),
  overwrite = TRUE
)
```

PR #65, opened 2025-06-20
Draw ui with state

nbenn:
@DivadNojnarg let me know what you think.

Side remark 1: Reactive in the state extracted from `rv$blocks$server$state` have to be "evaluated" before being passed to `expr_ui()`. Something like what I do in

https://github.com/BristolMyersSquibb/blockr.core/blob/918bff7111e3bd4f33d3a9d932c4492e9b22af4c/R/plugin-serialize.R#L93-L97

Side remark 2: All of this falls apart if you have UI in your block that is not part of the block "state". If this exists, it will be drawn with default values instead of "remembering". And this will happen, for example if your select options are something like `colnames()` of your input and not exposed via ctor.

I am not sure we can solve this issue with such an approach. Let's say you have a `selectInput()` and `choices` are some computed thing, such as `colnames()` of your input data. The current `choices` are lost on `removeUI()` and can only be recovered by running the `colnames()` calculation again. Does that make sense?

### DIFF
--- a/R/block-ui.R
+++ b/R/block-ui.R
@@ -41,14 +41,16 @@ block_ui.block <- function(id, x, ...) {
   NULL
 }
 
+#' @param state Optional block state, which will be added to the environment
+#' where the block `expr` UI is created
 #' @rdname block_ui
 #' @export
-expr_ui <- function(id, x, ...) {
+expr_ui <- function(id, x, state = list(), ...) {
   UseMethod("expr_ui", x)
 }
 
 #' @export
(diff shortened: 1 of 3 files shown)


DivadNojnarg:
@nbenn This does not work for me. Injecting the state does not trigger anything in the select block server function so the input stays empty, until I manually change the upstream data.

nbenn:
@DivadNojnarg have a look at `new_scatter_block()` for such an example.

PR #89, opened 2025-09-22
Syntax-highlight code

### LINKED ISSUE #44: Syntax highlight code
Do something like [pkgdown](https://github.com/r-lib/pkgdown/blob/de95b13f1ca3d09bee29f2f2742692bf30a28f8d/R/highlight.R#L3-L10).

nbenn:
Closes #44.


nbenn:
@DivadNojnarg I was experimenting with syntax-highlighting/auto-linking via downlit. One thing that I could not sort out though: For an app

```r
serve(
  new_board(
    c(a = new_dataset_block("iris"), b = new_scatter_block(x = "Sepal.Length", y = "Sepal.Width")),
    links = list(from = "a", to = "b", input = "data")
  )
)
```

If you show code and click on a link and then go back to the app, everything is greyed out (and stays that way), but the app is still functional. Do you know what is going on? Is that the same for you?

PR #101, opened 2025-10-07
Refactored board restore

### LINKED ISSUE #100: Refactor restore infrastructure
Currently, for a board "refresh" we "tear down" and "rebuild" UI:

https://github.com/BristolMyersSquibb/blockr.core/blob/f43de366e5da40cddb9ff5d23738cb59061ca6de/R/board-server.R#L243-L258

This is missing proper tear-down of server components and to make this more extensible adds considerable complexity: If we have module(s) created as part of "start" callback(s), reactive results have to be properly re-wired (to plugins). As consequence, depending on whether this allows for `reactive()` return values or is restricted to `reactiveVal()` + `observeEvent()`, this in turn requires re-initialization of plugins as well.

In short: having a cleaner "refresh" approach would be highly desirable.

Cc @DivadNojnarg

nbenn:
Fixes #100.


nbenn:
@DivadNojnarg currently this does not work:

- Some server-side issues I don't yet understand: might be something that's fixable or point to a more fundamental problem.
- The UI does not appear to be "re-rendered" using the new board on a reload.

Some work still necessary to figure out if this approach is viable.

nbenn:
Now the UI re-builld issue remains.

PR #105, opened 2025-10-26
Skip formatStyle for 0 col dataframes

### LINKED ISSUE #104: DAG boards break when blocks output 0x0 dataframes
- 0x0 dataframes lead to block output disappearing in DAG boards and won't restore
- Does NOT occur in regular new_board(), only new_dag_board() 

### Reproduction

Change the `select` choice in the example below to `FALSE`, to produce a 0x0 data.frame.

```r
  # Something weird with 0 x 0 dataframes. If we change the select selector
  # to FALSE, we get a 0 x 0 dataframe. This permanently messes up the workflow.
  # If I restore back to a normal value, it won't restore.
  blockr.core::serve(
    blockr.ui::new_dag_board(
      blocks = c(
        data = new_dataset_block(
          dataset = "iris"
        ),
        subset = new_subset_block(
          subset = "Species == FALSE", # 0 rows
          select = "Species"
        )
      ),
      links = list(
        new_link(from = "data", to = "subset")
      )
    )
  )

```
Interestingly, the problem does not exist in `blockr.core::new_board()`:

```r
  blockr.core::serve(
    blockr.core::new_board(
      blocks = c(
        data = new_dataset_block(
          dataset = "iris"
        ),
        subset = new_subset_block(
          subset = "Species == FALSE", # 0 rows
          select = "Species"
        )
      ),
      links = list(
        new_link(from = "data", to = "subset")
      )
    )
  )

```

Probable root cause:

- `DT::formatStyle(columns = character(0))` puts DT into broken state in DAG boards' dynamic UI.

christophsax:
The problem is very confusing (it occurs in the dag board but not in core), but the fix is simple and harmless.

Fixes #104

### DIFF
--- a/R/utils-dt.R
+++ b/R/utils-dt.R
@@ -41,15 +41,23 @@ dt_result <- function(result, session) {
   DT::renderDT(
     {
       dat <- as.data.frame(utils::head(result, rows))
-      DT::formatStyle(
-        DT::datatable(
-          dat,
-          selection = "none",
-          options = opts
-        ),
-        columns = names(dat),
-        whiteSpace = "pre-wrap"
+      dt <- DT::datatable(
+        dat,
+        selection = "none",
+        options = opts
       )
+
+      # Skip formatStyle for empty dataframes (0 columns)
+      # formatStyle with columns=character(0) breaks DT in DAG boards
+      if (ncol(dat) > 0) {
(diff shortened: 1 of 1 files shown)


nbenn (approved):
Lgtm, thanks!

PR #114, opened 2025-11-05
show non base data packages

christophsax:
we need this to read data from cards


nbenn (changes requested):
@christophsax I agree that having a pkg selector in the front end would be useful. The question is how to best implement this. Your current suggestion seems very arbitrary. There are some reasonable candidates in your list, but cards, gapminder, this just serves to fill our own case-specific needs.

What other options do we have? A text input that is validated against `rownames(installed.packages())`? Or go over all of `rownames(installed.packages())` and check for each whether `list_datasets(.) > 0`?

nbenn (commented):
One concern: I'm not sure here, but I think, this might eat up a bit of memory. You could ofc load into an environment and clean after retrieving the set of packages of interest. Then we don't pollute memory for the entire session. In memory constrained systems this might cause a bit of an issue though. Have you checked the order of magnitude of memory consumed by this "loading of all datasets"? Ofc this will depend on what packages are available in a given session.

christophsax:
I started something on 3 but it didn't go anywhere. We can do this another day.

nbenn:
@christophsax do we close here? Or is this still being worked on? I think it might make most sense to add a "fancy" dataset block to some place like blockr.io or so. One thing I like about the current `dataset_block` is that it also serves as an example of a block where we have a constructor argument that is runtime-static. I'll close. Please re-open if you disagree. Happy to work through the issues that came up if worthwhile.

christophsax:
fancy dataset block in blockr.io: good idea!

PR #134, opened 2026-01-09
External block control

### LINKED ISSUE #133: Control `expr` module state from block context
Expose `expr` reactive values responsible for block state to the block server context.

nbenn:
Fixes #133.

```deps
BristolMyersSquibb/blockr.dock#93
BristolMyersSquibb/blockr.dag#94
```


nbenn:
Open issue: we need to add additional information to the registry. How do we make this available to blocks created via direct `new_*_block()` calls?

We could in `new_block()` hook into the registry and try (if possible) to get blocks from there. Might need an additional argument to disable this in order to be able to break cycles.

nbenn:
@christophsax i have not been able to reproduce your issues. If i have a setup as

```r
serve(
  new_board(
    c(a = new_dataset_block(), b = new_subset_block()),
    links = list(from = "a", to = "b")),
    plugins = custom_plugins(ctrl_block()
  )
)
```

I can make valid, as well as invalid changes, such as subset: `Time > 5`, select: `c(Time, demand)`, or select: `c(Time, Demand)` and the checking logic flags this and suppresses down-stream eval.

For the subset block I did run into an issue with the whole bquoting we added late last year. This is now fixed, but I'm not sure this is what caused your issues.

Fyi, the whole `eval_impl()` should no longer be needed. The signature changed and you now just have a reactive `eval`, that you can, well, evaluate, e.g. in a `try()`/`tryCatch()`.

Let me know what needs to be done on my end to make progress here. If you have an (updated) example where this breaks, I'm happy to look into it.

PR #151, opened 2026-03-20
Fix deadlock in board reload

### LINKED ISSUE #147: Deadlock in board reload when loading workflow while reload is pending
### Problem

The while (is_reloading("reload")) { Sys.sleep(5) } loop in board-server.R causes a
deadlock when board_refresh fires while a session$reload() is already pending.

### How it happens

1. Load workflow A → update_serve_obj sets "reload" flag → session$reload() sends
reload to browser
2. Before browser reloads, a second board_refresh fires (user clicked Load on another
 workflow)
3. Observer sees is_reloading("reload") = TRUE → enters while loop → Sys.sleep(5)
blocks the R thread
4. finalize_reload() can never run (needs board_server() re-init, which needs R free)
5. Deadlock — stuck until Connect timeout (~10 min)

Easier to trigger on Connect due to session$reload() network latency.

### How to reproduce

1. Deploy on Connect
2. Load a workflow via URL or Load button
3. While reloading, quickly trigger another workflow load
4. App locks up with "Reload in progress." log messages

### Fix

Replace the blocking while with a non-blocking if — update the stored board and
return.

### Secondary concern

On Connect with multiple workers, .blockr_reload_state and serve_obj are
process-local. Cross-worker reloads lose the URL dedup state, causing extra reload
cycles.

nbenn:
## Summary

- Replace blocking `while`/`Sys.sleep` loop in the board refresh observer with cooperative waiting via `invalidateLater`, fixing the deadlock described in #147
- Serialize concurrent reloads across sessions using `is_reloading()` as a process-level gate, with R's single-threaded flush ordering providing mutual exclusion
- Add 60-second timeout to detect stale reload state (e.g. user closed browser mid-reload), clearing the stale entry and aborting the waiting session to prevent wrong-content leakage across sessions

Closes #147

### DIFF
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -293,37 +293,54 @@ board_server.board <- function(id, x, plugins = board_plugins(x),
       )
 
       if (not_null(board_refresh)) {
-        observeEvent(
-          board_refresh(),
-          {
-            while (is_reloading("reload")) {
-              notify(
-                "Reload in progress.",
-                duration = NULL,
-                id = session$token
-              )
-              Sys.sleep(5)
+
+        reload_pending <- reactiveVal(NULL)
+
+        observeEvent(board_refresh(), reload_pending(Sys.time()))
+
+        observe({
+
+          pending <- reload_pending()
+          req(pending)
+
+          if (is_reloading("reload")) {
+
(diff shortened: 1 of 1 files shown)


nbenn (commented):
LGTM.

PR #154, opened 2026-03-29
Add board preload API to fix double reload on refresh

### LINKED ISSUE #153: Fix double page load on browser refresh
When a board is loaded from a URL (e.g. `?board_name=Foo`) and the user refreshes the browser, the app performs two full page loads instead of one. The first load initializes with the default empty board, then the URL observer triggers `restore_board()` → `session$reload()`, causing a second load.

The fix adds a preload mechanism: `register_board_preload()` lets plugins register a callback that runs during the HTTP request phase (in `serve_board_ui`). When the callback returns a board, it's stored in `serve_obj` before either UI or server initialize, eliminating the redundant reload.

Changes:
- `serve_board_ui` now accepts `req` and calls a registered preload function when no reload is in progress
- New exported function `register_board_preload(fn)` to register the preload callback
- `update_serve_obj()` is now exported (was internal)

nbenn:
On browser refresh with a board URL (`?board_name=X`), the app previously performed two full page loads. The first load initialized with the default empty board, then the URL observer triggered `restore_board()` → `session$reload()` for a redundant second load.

The root cause: `serve_obj["reload"]` was empty on fresh page loads (no persisted state from `session$reload()`), so `serve_board_ui` and `serve_board_srv` both initialized with the wrong board.

The fix adds a preload hook that runs during the HTTP request phase — before either UI or server initialize. `serve_board_ui` now accepts `req` and calls a registered preload callback (guarded by `!is_reloading("reload")` to serialize concurrent reloads). When the callback returns a board, it populates `serve_obj["reload"]` so both UI and server start with the correct state.

New exports: `register_board_preload(fn)`, `update_serve_obj()`.

Closes #153


nbenn:
Closing in favor of #214. The board loader API there subsumes the request-phase preload: `register_board_preload` becomes the loader's cold-load path, so the double-load fix (#153) lands as part of the loader rather than a separate preload hook.

nbenn:
Superseded by #215, which landed the per-request board loader — the board for an incoming request is now resolved through a `board_loader` (default `local_loader()`), replacing the process-global serve-staging slot this PR worked around. Merged in 032c7a1.

PR #169, opened 2026-04-09
Lazy Eval

christophsax:
Do make sure blocks are not visible on startup, this worked in may tests:

https://github.com/cynkra/blockr.design/blob/main/open/lazy-eval/pages-visibility-requirement.md


nbenn:
Done.

PR #172, opened 2026-05-18
Inject code.namespace OTel attribute via otel::start_span trace

nbenn:
Reworks #171's per-call-site `otel_lbl()` decoration as automatic
package attribution via `code.namespace`, a standard OpenTelemetry code
semantic-convention attribute injected at span-start time.

A single `trace()` hook on `otel::start_span` reads the existing
`code.file.path` attribute that shiny already attaches via
`otel_srcref_attributes()`, resolves it to a loaded package using
`basename(dirname(dirname(path)))` validated against
`getNamespaceInfo(pkg, "path")`, and adds `code.namespace`. The
validated mapping is cached per package in a closure env embedded in
the tracer body via `bquote()` — no exported helper, no `:::` access.

This attribution flows through every reactive constructor uniformly —
`observe()`, `observeEvent()`, `reactive()`, `eventReactive()`,
`bindEvent()`, `bindCache()` — without depending on shiny's
`wrappedFunc` chain or the order in which it wraps observer/observable
internals, and without mutating any reactive object's slots. The trace
augments a formal argument that the function under trace was already
going to pass through to `start_span`.

Downstream packages (`blockr.dock`, `blockr.dag`) get attributed labels
in OTel traces with no code changes.

## Details

- `trace_start_span()` / `untrace_start_span()` / `is_start_span_traced()`
  follow the same pattern as the existing `trace_observe()` triple, with
  a separate `blockr.start_span_hook_disabled` option as the kill switch.
  Started alongside `trace_observe()` from `serve()` and from the test
  setup file.
- `otel` is in `Suggests`, not `Imports`. shiny only added OpenTelemetry
  in 1.12.1 while blockr.core targets `shiny (>= 1.5.0)`, so we can't
  rely on transitive availability. All three trace functions short-circuit
  via `pkg_avail("otel")` when otel isn't installed.
- The 83 explicit labels added by #171 stay as plain strings —
  descriptive labels are still useful — but the `otel_lbl()` helper
  itself, its export, man page, pkgdown entry and tests are removed.


nbenn:
@DivadNojnarg this is an alternative approach that covers package namespace information gathering for all kinds of otel traces without requiring per-package, per-observer/reactive/... changes.

The trace-based hook could be swapped out if we ever get otel-support for such hooks (https://github.com/r-lib/otelsdk/issues/77).

An important deviation to consider from your initial proposal: I'm attaching the package namespace data as an `code.namespace` attribute instead of pre-fixing the ID. This would allow us for easier filtering etc. But we can also prefix the ID. Or do both.

What do you think?

DivadNojnarg:
@nbenn as long as this information can be retrieved in the otel data column headers, I am fine.

PR #189, opened 2026-05-28
Add split() methods for blocks, links and stacks containers

### LINKED ISSUE #188: Add split() methods for container classes (blocks/links/stacks)
## Motivation

Iterating a container's elements as **wrapped one-element containers** (not bare elements) currently needs an anonymous lambda or an obscure trick. The recurring shape (from blockr.dock, where `block_panel()` takes a one-element `blocks` collection):

```r
lapply(blk_ids, function(id) block_panel(blocks[id]))
```

Because core's `[` preserves class, this already works today:

```r
split(blocks[blk_ids], seq_along(blk_ids))   # list of 1-element `blocks`, in order
```

…but it's verbose (explicit `seq_along`) and non-obvious as the idiom for "give me each element wrapped".

## Proposal

Add `split.blocks` / `split.links` / `split.stacks` (or one shared method) with an order-preserving singleton **default**:

```r
split.blocks <- function(x, f = seq_along(x), ...) NextMethod()
# split(blocks) -> list of 1-element `blocks`, input order preserved
```

The only value-add over base is the `f = seq_along(x)` default plus a documented, discoverable entry point. `split.default` already does the right thing via class-preserving `[`; an integer `f` gives numeric factor levels, so order is preserved (no lexical reshuffle).

## Open question: `split(x, ids)` as an ordered selector?

Suggested in review: a `split(blocks, ids)` that (a) does **not** reorder, (b) allows shorter/longer results (select a subset, or by id).

My concern is that this **conflicts with base `split` semantics**, so I'd keep it out of this change:

- In base R a character `f` is a **grouping factor** — parallel to `x` (`length(f) == length(x)`), coerced to a factor whose **levels are sorted**, partitioning every element into exactly one group. So `split(x, c("b","a"))` sorts to `a, b` and rejects a length mismatch.
- Re-interpreting a character `f` as an **ordered, length-flexible selector** would silently break those expectations for anyone who knows `split`. `split(x, ids)` looking like grouping but behaving like `x[ids] |> singletonize` is a footgun.

If the "select-by-id, ordered, flexible-length" operation is worth a verb, I'd give it its own name rather than overload `split`'s character-`f` path — or just keep the explicit two-step `split(x[ids], seq_along(ids))` (subset, then singletonize), which says exactly what it does.

So: scope this issue to the `f = seq_along(x)` default-singleton methods; treat the ordered-selector idea as a separate design question.

## Out of scope

blockr.dock's `dock_extensions` / `dock_layouts` are plain-list-backed — their `[` returns a bare `list`, so `split` doesn't wrap them. Bringing those in line (class-preserving `[`, or their own `split`) is a blockr.dock concern, tracked there.

nbenn:
## Summary

- Add `split()` methods for the `blocks`, `links` and `stacks` container classes. The grouping `f` defaults to `seq_along(x)`, so `split(x)` returns each element wrapped in a one-element container of the same class, in input order — the discoverable idiom for iterating a container as singletons rather than via the obscure `split(x, seq_along(x))`.
- The methods delegate to `split.default(x, f, ...)` rather than the issue's proposed `NextMethod()` one-liner. `NextMethod()` reconstructs the call from the generic's matched arguments, so a *defaulted* `f` is never passed on; since `split.default()` has no default for `f`, the bare-`NextMethod()` form errors with `argument "f" is missing` on `split(x)`. There is no intermediate `split` method (no vctrs method exists) to chain to, so delegating to `split.default()` skips nothing.
- Scoped to the singleton-default methods, as the issue requests: no character-`f` "ordered selector" overload, so standard `base::split()` grouping semantics are preserved when `f` is supplied.
- Adds a shared `?split.blocks` help page and tests covering the singleton default, order preservation (incl. >10 elements, where an integer `f` avoids lexical reshuffle), empty containers, and explicit-`f` grouping.

Fixes #188


nbenn:
Closing — we're not convinced the split() container methods are worth adding at this point. Can revisit if a concrete need comes up.

PR #209, opened 2026-06-04
Key the board reload handoff per tab, not a global slot

### LINKED ISSUE #208: Per-tab reload handoff via a sessionStorage tab id
## Problem

`serve()` stages the next board for a reload in a single process-global slot, `serve_obj$reload` (`utils-serve.R:243`). `update_serve_obj("reload", board)` writes it just before `session$reload()` (`board-server.R:283`), and the returning session consumes it unconditionally at `board_server()` startup via `finalize_reload("reload")` (`board-server.R:38`).

The slot is keyed by the literal string `"reload"`, not by session — there is no stable id across the `session$reload()` boundary. Under one shared process (Posit Connect, multiple users), "the returning session" is just whoever boots `board_server()` next, which need not be the session the board was staged for.

The `is_reloading()` gate (`board-server.R:250`) is a write-side mutex: it stops two sessions from staging at once, but it does not — cannot — bind the consumer to the writer. A fresh visitor never passes through the gate yet still consumes the staged slot at line 38.

This strands state two ways, both reproduced/traced:

- **Board contamination** — an unrelated session renders a board staged by someone else (reproduced deterministically; see Validation).
- **The reload loop** — the slot also carries `meta`, which blockr.session's anti-loop guard depends on (`prev_query` is seeded from `board$reload_meta$url`; blockr.session `server.R:27`, `76-78`). When that meta is stranded under concurrency, the guard misfires and re-restores, producing chained reloads that terminate after a variable-but-finite count once the contention quiesces. This matches the long-standing field reports.

## Why not drop the reload

The reload is deliberate and worth keeping: it avoids reliance on (still-unreleased, non-API) module teardown, gives a genuine clean slate, and can swap the whole app (e.g. core to dock), not just the board. So the fix targets the handoff, not the reload.

## Proposal

Give each browser tab a stable id in `sessionStorage` (per-tab, survives reload, cleared on tab close, never shared across tabs) and key the staged state server-side by that id instead of the single `"reload"` slot. Each returning session then picks up exactly its own staged board. This is the root fix for both symptoms, and it lets the global `is_reloading` gate and the 60s stale-close (`board-server.R:250-267`) retire, since per-tab keys do not contend.

The one constraint: `sessionStorage` is unreadable at the HTTP GET that builds the UI — no client JS has run yet. But the reload is server-initiated, so the reloading session already holds its tab id (received post-connect) and can carry it across. Three ways to bridge, in order of preference:

1. **Reload URL (start here)** — before `session$reload()`, stamp a one-time handoff token into the URL (`updateQueryString(..., "replace")`); the post-reload UI reads it from the request (`serve_board_ui(req)` then `parseQueryString(req$QUERY_STRING)`) and resolves the staged board synchronously. Preserves today's synchronous full-page render. Builds on the request-phase resolution #153 introduces (`serve_board_ui(req)`, exported `update_serve_obj`), extending it so the reload handoff resolves per tab via a token in the request.
2. **Bounce page** — the first GET returns only JS that reads the `sessionStorage` id and redirects with it in the URL, then renders. A wasted round-trip; fallback only.
3. **Deferred render** — render a neutral shell at GET, hydrate the board via `renderUI` once the id arrives post-connect. Makes `sessionStorage` sufficient with no URL token, but turns the board back into dynamic UI, against the grain of why we reload. Last resort.

`sessionStorage` stays the canonical per-tab id (read once, post-connect); the URL token is a disposable per-reload bridge that can be stripped after pickup, so the persistent id need not linger in the address bar.

## Relationship to #153

#153 eliminates the double page load on refresh by resolving the board during the HTTP request phase, but still through the shared `serve_obj` — so it is single-session and does not address cross-session contamination. This issue adds per-tab keying on top: #153 decides "resolve at request phase", this decides "resolve the right board for this tab". Approach 1 rides #153's request-phase plumbing directly, so they should land together.

## Validation

Multi-session / single-process tests — the bug only manifests with concurrent sessions in one process. The harness (already prototyped): start the app with shinytest2 (`AppDriver$new(app_dir)`), then attach a second session to the same process by URL (`AppDriver$new(owner$get_url())`). Assert DOM facts via xml2, not snapshots:

- An unrelated session renders its own board, never one staged by another session.
- A session's own staged board survives a concurrent fresh connection (no stranding).

The downstream loop (blockr.session) is validated separately by driving its query-restore on the same harness and asserting no chained reloads.

## Acceptance

- Staged reload state is keyed per browser tab, not by a single shared `"reload"` slot.
- Two-session / one-process test: the bystander renders its own board; the owner's staged board is not stranded.
- The anti-loop marker no longer depends on the process-global slot meta.
- Approach 1 landed; approaches 2 and 3 documented as fallbacks if it does not pan out.

nbenn:
## Summary

- A `session$reload()` board restore staged the next board in a single process-global slot (`serve_obj$reload`), consumed unconditionally by whichever session booted `board_server()` next. With no stable id across the reload boundary, under one shared process (e.g. Posit Connect) an unrelated session could render or strand a board staged for someone else — the cross-session contamination, and the chained-reload loop when the stranded slot's `meta` misfired blockr.session's anti-loop guard.
- The board is now staged under a one-time, per-reload token. `stage_reload_handoff()` mints a token, stages it under `reload-<token>`, stamps `?__blockr_reload__=<token>` into the URL (additive `updateQueryString(..., "replace")`) and reloads. Both the request-phase UI (from `req$QUERY_STRING`) and the server (from `session$clientData$url_search`, available synchronously) resolve it via a shared `staged_board_key()`, so each returning session picks up exactly its own board; the token is stripped from the address bar after the consume. The request-phase preload from #153 moves to a per-query key in the same step, so no path reads a shared slot.
- Stale handoff / preload slots (cross-process reloads, double restores, tabs closed mid-reload) are swept on a TTL when a new slot is staged — this replaces the global `is_reloading` gate and 60s stale-close the PR removes.

This is approach 1 from the issue (the one-time URL handoff token). I did not add the sessionStorage tab id: `sessionStorage` is unreadable at the GET that builds the UI, so the URL token carries the handoff regardless, and a server-minted per-reload token already gives each returning session its own slot. The server reads `session$clientData$url_search` synchronously at the top of the server function, so the board resolves at request time with no bounce page or deferred render (approaches 2/3). Glad to layer a stable per-tab id on top if you want one for something beyond this handoff.

### URL-param contract with blockr.session

The token write here is additive — a read-modify-write that preserves other params, with the token namespaced as `__blockr_reload__`. blockr.session, by contrast, writes the URL destructively: `board_query_string()` rebuilds the query from `board_name`/`user`/`version` only, and `manage_project_server`'s `new_btn` handler clears it to `"?"` (`server.R:149`), so it would drop `__blockr_reload__`. This is safe today only by ordering — blockr.session sets `board_name` and fires the restore, then this producer appends the token and reloads with no blockr.session write in between — and by the token being ephemeral (consumed and stripped on the next load). Tracked as a follow-up to make the contract robust (additive writes, or a reserved `__blockr_*` namespace every writer preserves): BristolMyersSquibb/blockr.session#53.

The loop fix here is the *root cause* — correct per-tab `meta` delivery. It is verified by reading `manage_project_server`'s token-insensitive guard (it rebuilds the canonical URL from `board_name`/`user`/`version` and compares that to `prev_query`, so `__blockr_reload__` cannot trigger a spurious restore); end-to-end loop validation belongs downstream in blockr.session and is noted in the follow-up.

Stacked on #153 (base `153-double-load`); it extends that PR's request-phase resolution and they land together. CI gates on a `main` base, so checks stay idle until the base retargets to `main` after #153 merges — validated locally meanwhile (full suite green; the repro file is proven red on the unfixed base and green with the fix).

Fixes #208

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,12 @@
 # blockr.core 0.1.3
 
+* The board reload handoff (the `session$reload()` board restore) is keyed per
+  reload rather than via a single process-global slot: the board is staged
+  under a one-time token that rides across the reload in the URL and is
+  resolved per request. Under one shared process (e.g. Posit Connect) a session
+  no longer renders or strands a board staged for another — the root cause of
+  the cross-session board contamination and the associated reload loop. Stale
+  handoff slots are evicted on a TTL (#208).
 * New exported generic `external_ctrl_vars()` (with a `block` method) and
   predicate `has_external_ctrl()` provide a public, polymorphic API for
   resolving a board component's `external_ctrl` declaration into the set of
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -35,7 +35,7 @@ board_server.board <- function(id, x, plugins = board_plugins(x),
                                callback_location = c("end", "start"),
                                ...) {
 
-  reload_meta <- finalize_reload("reload")
+  reload_meta <- finalize_reload(staged_board_key(session_url_query()))
 
   plugins <- as_plugins(plugins)
 
@@ -238,53 +238,22 @@ board_server.board <- function(id, x, plugins = board_plugins(x),
 
       if (not_null(board_refresh)) {
 
-        reload_pending <- reactiveVal(NULL)
+        observeEvent(
+          board_refresh(),
+          {
+            val <- board_refresh()
 
-        observeEvent(board_refresh(), reload_pending(Sys.time()))
-
-        observe({
-
-          pending <- reload_pending()
-          req(pending)
-
-          if (is_reloading("reload")) {
-
-            if (difftime(Sys.time(), pending, units = "secs") >= 60) {
-              log_warn("stale reload state, clearing")
-              finalize_reload("reload")
-              reload_pending(NULL)
-              session$close()
-              return()
+            if (is_board(val)) {
+              board <- val
+              meta <- NULL
+            } else {
+              board <- val$board
+              meta <- val$meta
             }
 
-            notify(
-              "Reload in progress.",
-              duration = NULL,
-              id = session$token
-            )
-            invalidateLater(1000)
-            return()
+            stage_reload_handoff(board, meta, session)
           }
-
-          removeNotification(session$token)
-          reload_pending(NULL)
-
-          val <- isolate(board_refresh())
-
-          if (is_board(val)) {
-            board <- val
-            meta <- NULL
-          } else {
-            board <- val$board
-            meta <- val$meta
-          }
-
-          log_debug("refreshing board")
-          update_serve_obj("reload", board, meta = meta)
-
-          log_debug("reloading session")
-          session$reload()
-        })
+        )
       }
 
       call_plugin_server(
--- a/R/utils-serve.R
+++ b/R/utils-serve.R
@@ -206,13 +206,14 @@ serve_board_ui <- function(id, plugins, options, ...) {
 
   function(req) {
 
-    if (!is_reloading("reload")) {
+    query <- parseQueryString(coal(req$QUERY_STRING, ""))
+    key <- staged_board_key(query)
+
+    if (!is_reloading(key)) {
 
       preload <- get0("preload_fn", envir = serve_obj, inherits = FALSE)
 
       if (not_null(preload)) {
-
-        query <- parseQueryString(coal(req$QUERY_STRING, ""))
 
         result <- validate_preload_result(
           tryCatch(
@@ -225,12 +226,12 @@ serve_board_ui <- function(id, plugins, options, ...) {
         )
 
         if (not_null(result)) {
-          update_serve_obj("reload", result$board, meta = result$meta)
+          update_serve_obj(key, result$board, meta = result$meta)
         }
       }
     }
 
-    x <- get_serve_obj("reload")
+    x <- get_serve_obj(key)
     id <- coal(attr(x, "id"), id)
 
     log_debug("building ui for board {id}")
@@ -250,13 +251,28 @@ serve_board_srv <- function(id, plugins, options, ...) {
 
     onStop(revert(trace_observe()), session)
 
-    x <- get_serve_obj("reload")
+    query <- session_url_query(session)
+    x <- get_serve_obj(staged_board_key(query))
     id <- coal(attr(x, "id"), id)
 
     res <- do.call(
       blockr_app_server,
       c(list(id, x, plugins = plugins(x), options = options(x)), args)
     )
+
+    # `board_server()` consumes the staged slot synchronously above (its
+    # `finalize_reload()` runs during the call). Strip the one-time handoff
+    # token from the address bar only afterwards, so the consume provably
+    # reads the live token rather than racing the (asynchronous, client-side)
+    # `updateQueryString()`.
+    if (not_null(reload_token(query))) {
+      query[[reload_query_param]] <- NULL
+      updateQueryString(
+        query_to_string(query),
+        mode = "replace",
+        session = session
+      )
+    }
 
     blockr_test_exports(x, res)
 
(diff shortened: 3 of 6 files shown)


nbenn (commented):
Design-review follow-up (from the broader approach discussion). The approach is right and most of it is solid — these are the changes I'd want before it lands.

Context: the base PR #154 is now rebased onto `main` (0.1.3) and CI-green, so this branch needs rebasing onto the new `153-double-load` (its old base tip was rewritten).

## Must-fix

1. **Orphaned `reload-<token>` slots are never evicted — a memory leak.** The `board_refresh` observer in `R/board-server.R` mints a unique key per restore (`update_serve_obj(paste0("reload-", token), ...)`), freed only when the matching session returns and consumes it. Cross-process reloads (Connect affinity miss), double-restores (token B overwrites the URL, token A's slot leaks), and tab-close-during-reload leave full board objects in the process-global `serve_obj` for the lifetime of the process — and this PR removed the 60s stale-close that was the only cleanup. This is a regression versus the old single-slot design. Add a TTL/bounded sweep: stamp each staged slot and evict expired ones on stage/access.

## Should-fix

2. **The e2e no longer exercises the new mechanism.** The fixture (`tests/testthat/apps/reload-repro/app.R`) still plants the literal `"reload"` slot, which nothing reads any more, so "an unrelated session does not render a board staged by another" passes because the key is dead — not because the per-token handoff isolates. Rename it to a "legacy shared slot is no longer read" guard, and add a positive e2e that drives the real path: one session does an actual restore (mint token, `updateQueryString`, reload, lands on its own board) while a concurrent fresh session is unaffected.

3. **Strip-vs-consume ordering is a load-bearing, undocumented assumption.** `serve_board_srv` strips the token (async `updateQueryString`) before `board_server` consumes the slot by re-reading `clientData$url_search`; this is correct only because `updateQueryString` does not synchronously update server-side `clientData`. Move the strip to after `do.call(blockr_app_server, ...)` so the consume provably runs first, and/or add a comment making the assumption explicit.

## Cross-package coordination (URL params)

4. The additive read-modify-write here (preserve other params, namespace the token as `__blockr_reload__`) is the right pattern. But this shares the un-namespaced URL-param global with blockr.session, which writes destructively — `board_query_string` rebuilds from `board_name`/`user`/`version` only, and `server.R:149` clears the query to `"?"` — dropping `__blockr_reload__`. It works today only because of ordering (blockr.session writes board_name, fires the restore, then this producer appends the token and reloads with no blockr.session write in between) and the token being ephemeral. Please document this contract in the PR, and open a follow-up: either make blockr.session's URL writes additive (preserve unknown params) or reserve a `__blockr_*` namespace that every writer must preserve.

## Minor

5. NEWS: the `# blockr.core 0.1.2` header collides with the now-0.1.3 base on rebase; reconcile it, and soften "fixing the reload loop" to "the root cause of the loop" until it is validated downstream in blockr.session — the loop fix is currently verified by reading `manage_project_server`, not by running it, so add that test there. `query_to_string` does not URL-encode param names (safe for the current `board_name`/`user`/`version`/token set, but technically incomplete). The `id = "reload"` defaults on `is_reloading`/`finalize_reload` are now vestigial.

nbenn:
Thanks for the thorough review — all addressed, and rebased onto the reworked `153-double-load` (0.1.3).

1. **Orphaned slots / leak.** Added a TTL sweep: `update_serve_obj()` stamps each staged slot, and staging sweeps `reload-*` / `preload-*` entries older than `reload_handoff_ttl` (600s). Unit-tested — a stale slot is evicted on the next stage, while a fresh slot and `initial` are kept.

2. **E2e.** Renamed the old test to "a legacy shared 'reload' slot is never read by a fresh session" — it now explicitly guards the dead global slot. Added a positive e2e ("a returning session lands on its own board, a fresh one does not"): the returning session arrives with its token in the URL and renders its staged board, while a concurrent token-less session renders the initial board. I drove the producer (mint → stage → meta) in a deterministic unit test on `stage_reload_handoff()` rather than via a mid-test `session$reload()`, because a chromote `Page.navigate` mid-test is the known-flaky operation that ejects the merge queue here. Happy to add the full reload-in-browser variant if you'd rather have it despite that.

3. **Strip ordering.** Moved the strip to after `do.call(blockr_app_server, ...)`, so `board_server()`'s `finalize_reload()` provably consumes the slot first; added a comment stating the assumption.

4. **URL-param contract.** Documented in the PR body, and opened BristolMyersSquibb/blockr.session#53 to make blockr.session's URL writes additive or reserve a `__blockr_*` namespace.

5. **Minor.** NEWS entry moved under 0.1.3 and softened to "the root cause of the … loop"; `query_to_string()` now URL-encodes param names too; dropped the vestigial `"reload"` defaults on `is_reloading()` / `finalize_reload()`. The end-to-end loop test belongs downstream in blockr.session — tracked as BristolMyersSquibb/blockr.session#54.

nbenn:
Superseded by #214. Per-tab keying makes the shared `serve_obj` slot safe; #214 removes the slot entirely — board resolution moves to a request-phase `loader` on the `preserve_board` plugin — so cross-session contamination becomes impossible by construction rather than keyed-safe. Closing in favor of #214.

nbenn:
Superseded by #215, which landed the per-request board loader — the board for an incoming request is now resolved through a `board_loader` (default `local_loader()`), replacing the process-global serve-staging slot this PR worked around. Merged in 032c7a1.

Issue #203, opened 2026-06-04
Gate block evaluation and rendering on visibility


nbenn:
A large board evaluates every block and renders every output at startup, and on every update fans the recompute out through the whole reachable subgraph — all of it regardless of what the user can actually see. This adds a visibility gate so a board only computes what's on screen and what feeds it.

## Approach

Add a board-level write channel, `visible`, carrying the ids of the blocks currently on screen. `board_server()` hands it to the board callback alongside `update` (same shape as the existing `update` channel), so a front-end such as blockr.dock can drive it. When nothing writes it, every block is treated as visible and behaviour is unchanged from today.

Core owns the dependency reasoning. A block **renders** only while it is itself in `visible`. A block **evaluates** while it is in `visible` or is an upstream dependency of a visible block — an off-screen intermediate still feeds its visible descendants, but does no rendering. Both conditions are derived from the single `visible` set, the eval condition by walking `board_links` upstream; the producer never reasons about betweenness.

Each block's data-evaluation and output-render observers are constructed **suspended** and resumed by one small controller `observe` inside the block server — gating eval on the visible-or-upstream condition and render on plain visibility. The two observer helpers already return their `observeEvent` handles (`block-server.R:359-378`, `:407-424`), so this is: pass each block its gates, capture the handles, toggle them. Starting them suspended at init is what avoids evaluating and rendering off-screen blocks on first load; resuming re-runs them to catch up on reveal.

A `blockr_option` master switch turns the whole thing off.

## Scope

In: blocks. Suspend/resume as the gating mechanism. The `visible` channel, the upstream closure for the eval condition, the per-block render and eval gates, the option switch.

Out: extensions and lazy view-construction (front-end side, later); deferred block-server construction (parked — pursued only if construction proves slow once evaluation is gated); converting the eval/link path to lazy pull (separate follow-up).

## Done when

- With a producer driving `visible`, off-screen blocks neither evaluate nor render at init, and an update that doesn't reach the visible subtree triggers no work in it.
- Revealing a previously off-screen block evaluates and renders it.
- An off-screen block that is an upstream dependency of a visible block still evaluates (so its visible descendants update) but does not render.
- With no producer (channel unset), startup and update behaviour is unchanged from today; the full test suite is unaffected.
- The option switch disables gating entirely.

---

Producer side: BristolMyersSquibb/blockr.dock#195. Follow-up: #204.

PR #211, opened 2026-06-05
Make block evaluation demand-driven via lazy pull

### LINKED ISSUE #204: Demand-driven block evaluation (observer → lazy pull)
Follow-up to #203. Stage 1 (that issue) gates visibility by suspending observers and keeping an explicit upstream closure alive for the eval condition. This converts the evaluation path to lazy pull so the closure becomes implicit and an off-screen subtree goes fully quiescent — including the cheap `validate` / `state_check` observers that stage 1 leaves ticking — with no gating code at all.

## Approach

Turn the eager evaluation step (`data_eval_observer`, `block-server.R:362`) and the eager cross-block propagation (`setup_link`, `board-server.R:456`) into demand-driven reactives, so a block computes only when a visible leaf pulls it and the reactive graph computes the dependency closure itself. The per-block `visible` gate from #203 stays as the leaf-level demand signal; what changes is how that demand reaches upstream — from suspend-plus-closure to pure pull.

## Why it's separate

It touches the central link/input model — variadic `...args` inputs, `ignoreNULL = FALSE` propagation semantics, dynamic link add/remove — so it carries real regression risk and should land on top of a working #203 and be measured against it. It is not required for the stage-1 responsiveness win.

## Done when

- A block with no visible descendant receives no input invalidation and runs none of its reactivity (eval, render, validate, state-check) until something visible demands it.
- The explicit upstream closure from #203 is no longer needed to keep intermediate blocks live; the reactive graph does it.
- No behavioural regression in link propagation across the existing test suite.

Builds on #203.

nbenn:
## Summary

- A block's result becomes a lazy `reactive()` and downstream blocks pull their upstream results, replacing the per-link `observeEvent`s that eagerly pushed each upstream result into the downstream input. Link wiring is now a per-input source holder that the pull reactive reads, so add/remove stays dynamic.
- An off-screen block with no on-screen descendant is never pulled, so it runs no evaluation, validation or state-check until something visible demands it — the reactive graph computes the upstream closure on its own, which is the stage-2 quiescence win this issue is about.
- Drops the explicit `upstream_blocks()` closure and the suspended eval observer from #203; only rendering stays gated on visibility (render observer suspend/resume), and eval/validate/state-check fold into the pulled result chain with condition capture isolated so the `cond` writes don't self-invalidate.
- Preserves the existing `result()` contract: it is NULL-or-value and never throws. An unready block (gate closed, state unset, or invalid input data) returns NULL exactly as before, so link inputs and the edit-block summary keep seeing NULL rather than a silent `req()` error.

Stacked on #206 — base is `203-visibility-gating`, so this targets that branch until #206 lands, then retargets to `main`.

Fixes #204


christophsax:
### Test this against the real cdex app before merge — minimal boards miss the gating bug

This PR is the **gate** side of the visibility-gating empty-render bug filed as
[blockr.dock#254](https://github.com/BristolMyersSquibb/blockr.dock/issues/254): during
relayout the dock's `visible` producer ([blockr.dock#201](https://github.com/BristolMyersSquibb/blockr.dock/pull/201))
can emit a **terminal `0/N on screen`** report while the active view's panels *are* on
screen. Core gating here treats that trailing `0` as authoritative, suspends every block,
and never re-resumes once the panels settle → every view renders empty.

**The lesson worth encoding:** #254 passed *every* minimal-board test — a 2-view gated
fixture and synthetic boards of 11 views × 5 head-blocks all rendered fine. It only
reproduced on the **real cdex app**, because it's a timing race that needs real (slow)
blocks and a heavy active view (≈≥9 panels) to make the spurious `0/N` land **last**.
Minimal/synthetic boards are necessary but **not sufficient** coverage for this path.

**Asks for this PR:**

1. **Gate fix:** core must not treat a `0/N` as authoritative with no recovery — when the
   active view's panels settle visible, gating must re-resume them. (Fixing either side
   closes the bug; fixing both makes the contract robust.)
2. **Regression test (preferred, durable):** a headless shinytest2/chromote test that
   boots a heavy multi-panel board under `options(blockr.gate_visibility = TRUE)` and
   asserts the **active** dock renders after settle:
   ```js
   document.querySelector('.blockr-view-dock-active')
           .querySelectorAll('.dv-watermark,[class*=watermark]').length === 0
   ```
   (equivalently: the settled `visibility update` is not a terminal `0/N`). A
   self-contained fixture works — one view with ~10 real chart/table blocks reproduces
   it; no full cdex dependency needed. This is the test that would have caught #254.
3. **PR checklist:** *"smoke-tested against the real cdex app (or a heavy single view via
   `cedx_view_board("Lab")`) with `blockr.gate_visibility = TRUE` — active view has
   panels, no terminal `0/N` report."*

Repro harness + bisection + instrumented timeline are in #254.

nbenn:
## Quiescence has a hole: expr-server observers sit outside core's suspension surface

Scope note: this is the data-input slice of the leak, which folds into this PR. The external-control slice of the same problem, and the residual triggers, are tracked in #240.

Core's suspension surface is one observer per block — the render observer. Every block's `expr_server()` registers its *own* observers, whose handles core never receives, and the `data` list handed to the expr server is the **same** list of `upstream_result()` pull-reactives that `res` reads (`setup_block()` passes `rv$inputs[[id]]` to both `block_server` and `expr_server`).

So the bread-and-butter "populate my control from the incoming data" pattern is an **eager** observer reading a pull-reactive:

```r
# head_block here; ~every blockr.dplyr transform block (select, filter, mutate, arrange, summarize, ...)
observeEvent(data(), { ... update*Input() / sendCustomMessage() ... })
```

It runs at the first flush, reads `data()`, which calls `rv$blocks[[from]]$server$result()` and **pulls the whole upstream closure off screen**. Lazy `res` never fires, but the author's observer reaches around it and forces the cascade anyway. So #204's "receives no input invalidation and runs none of its reactivity" does not hold for any board containing a data-observing block — `head_block` plus essentially all of blockr.dplyr and blockr.ggplot. The regression is specific to lazy pull: under #206 this same observer only burned wasted UI work, because eager links meant reading `data()` didn't pull. Lazy pull is what upgrades it from "wasted work" to "defeats the mechanism."

## Fix: gate the shared `data` reactives on a `needed` signal

Reintroduce the per-block `needed` set this series already reasoned about — `visible` plus its upstream closure over `board_links()`, i.e. the `rv$eval` set #206 computed and #211 dropped (a cheap structural walk, recomputed only when `visible` or the links change). Then gate the pull-reactive on it:

```r
upstream_result <- function(src, rv, needed) {
  reactive({
    if (!needed()) return(NULL)
    from <- reval_if(src)
    if (is.null(from)) NULL else rv$blocks[[from]]$server$result()
  })
}
```

Off screen, `data()` is NULL, the eager observer's `ignoreNULL` skips its body, and nothing pulls upstream; on reveal `needed()` flips, `data()` invalidates and pulls once to catch up. Both the expr server and the core eval path read these same gated reactives, so the pull genuinely stops at the gate for the whole block — which is what makes the quiescence claim actually true rather than true-modulo-the-expr-server. Contract change to document: `data()` reads NULL while a block is neither visible nor feeding a visible block.

The `needed` set this introduces is also what the external-control follow-up (#240) gates on, so it is worth landing here in a shape that work can reuse.

nbenn:
Pushed `e4015f7`, which fixes an over-evaluation I found while we were tracing the reactive shape here. It changes `needed`, so it's worth a fresh look.

Evaluation was transitively gated on the whole board object: `res → dat_eval → dat → upstream_result → rv$needed() → rv$board`. Because `rv$needed` was a plain `reactive()` reading `rv$board`, and Shiny propagates reactive invalidation eagerly — regardless of whether the recomputed value differs — *any* `rv$board` reassignment (a cosmetic block rename, a stack edit, an unrelated link change) re-invalidated every needed non-source block's input and re-ran `eval_impl` on the next pull, even with byte-identical data. On a large gated board that's an O(needed) re-eval on essentially every structural edit. Block parameter/state edits were already correctly scoped — they go through the per-block state reactiveVals, not `rv$board`.

The fix holds `needed` in a `reactiveVal` maintained by an `observe` that writes only when the set actually changes (`setequal`, with the all-visible `TRUE` sentinel compared via `identical`). A `reactiveVal` suppresses notification on an unchanged set, which a `reactive()` cannot, so structural edits that leave the needed set intact no longer re-evaluate anything.

Regression test `an unrelated structural edit does not re-evaluate needed blocks`: under an active producer it renames an unrelated block and asserts a needed block does not re-evaluate. It fails before this commit and passes after. Full suite green including the app tests; no doc changes.

For context: this is the *gating* axis. The per-edge `rv$sources` routing holders stay as-is. I did check whether they could be retired by reading link targets from `board_links(rv$board)` instead, but that reintroduces the same whole-board dependency on the routing path — value-stability only suppresses propagation at a `reactiveVal`, and the per-edge routing target needs one such node per edge, which is precisely what `rv$sources` already is.

nbenn:
## Named variadic inputs don't render in the manage-links editor

`bb319a8` ("Stop the link editor from suggesting integer input names") over-trimmed `dt_board_link()`: it dropped the link's *current input names* along with the integer suggestions, so a variadic link whose input is a **name** no longer shows up in the editor.

Before `bb319a8`, a variadic target's `in_avail` was `union(seq_len-integers, cur_inp)` — the `cur_inp` half carried the names already assigned to that target's links. The commit replaced the whole expression with an empty list:

```r
in_avail <- c(
  lapply(blks[!is.na(arity)], block_inputs),
  set_names(
    rep(list(character()), sum(is.na(arity))),
    names(arity)[is.na(arity)]
  )
)
```

So a link with `input = "left"` has `"left"` *absent* from its choice list. `selectizeInput()` only emits `<option>` tags for values in `choices` — a `selected` value that isn't a choice is dropped server-side — so nothing renders as selected. Typing a fresh name via `create` then disappears on the next `DT` redraw, which rebuilds the cell from the (empty) choices.

### Repro

```r
b <- new_board(
  blocks = c(a = new_dataset_block("BOD"), b = new_dataset_block("BOD"), c = new_rbind_block()),
  links = links(l1 = new_link("a", "c", "left"))
)
dt <- dt_board_link(board_links(b), NS("x"), b)
# dt$Input[[1L]] carries only <option value=""></option>; "left" is absent.
```

### Fix

Keep the current input names in the variadic targets' `in_avail`; drop only the integer suggestions. Sketch:

```r
cur_inp <- split(lnk$input, lnk$to)
in_avail <- c(
  lapply(blks[!is.na(arity)], block_inputs),
  cur_inp[intersect(names(cur_inp), names(arity)[is.na(arity)])]
)
```

With `"left"` back in the choices the cell renders `<option value="left" selected>left</option>`; positional (empty) links still render empty; and no integer ever appears (the names are user-typed strings under the new model), so the intent of `bb319a8` holds.

The sibling spot in `create_dt_link_obs()` — the `to`-change branch that now sets `inp <- character()` — has the same shape: a retargeted link that already carries a name should keep that name in `choices` there too, otherwise re-selecting the row drops it.

### Downstream

blockr.dock#276 (per blockr.dock#275) mirrors this editor logic — `dt_board_link()` and the observer — so it carries the identical bug. I'll re-mirror whatever lands here.

nbenn:
Fixed in `a7c948e`.

Your diagnosis was exactly right — dropping the whole variadic `in_avail` took the `cur_inp` names with it. Restored the current input names for variadic targets (dropping only the integer suggestions), and the sibling `to`-change observer now keeps the retargeted link's current input in its choices too, so re-selecting the row no longer drops a name.

One small deviation from your sketch: I filter the positional empty out of `cur_inp` (`lapply(split(lnk$input, lnk$to), filter_empty)`), so a positional link renders a single `<option value="" selected>` rather than two — the un-filtered version emitted a doubled empty option (two `selected` empties). Named links are unaffected: a link with `input = "left"` now renders `<option value="left" selected>`, and no integer is ever offered.

Regression test `a named variadic link renders its name in the editor` asserts `value="left" selected`; it fails on `bb319a8` and passes now. The prior test only exercised positional (empty) inputs, which is exactly why it slipped through.

Full suite green, app tests included (after reinstall). Worth folding the `filter_empty` tweak into the dock#276 mirror.

nbenn:
## Second duplicated editor bug: clearing a variadic input name can't be applied

Related to the rendering issue above, in the same `create_dt_link_obs()` code that blockr.dock mirrors. The shared link-cell observer guards with:

```r
if (new == cur || new == "") {
  return()
}
```

Skipping every empty value is right for `from`/`to` (empty means "nothing selected"), but wrong for the `input` column: under the name-or-position model an empty input is a *valid positional slot*. So clearing a name (named → positional) returns early and never stages — the manage-links Apply button stays disabled and there's no way to commit the change. (The `identical(new, "")` branch in the `to` handler is also unreachable for the same reason, though that's a separate, older matter.)

### Fix

Let an empty value through for the input column; keep skipping it for from/to:

```r
if (new == cur || (new == "" && col != "input")) {
  return()
}
```

### Downstream

blockr.dock#276 fixes its duplicated copy independently (the code isn't shared), with a regression test that clears a named variadic link and asserts the edit stages. Same fix applies to this copy.

nbenn:
## Editing a link moves it to the end — changes positional order

Editing an existing link in the manage-links editor (its input name, from, or to) moves it to the end of the link order. Under the name-or-position model that silently changes positional argument order — e.g. editing the first of two positional links into an `rbind` turns `rbind(a, b)` into `rbind(b, a)`.

### Root cause

`edit_link_observer()` stages an edit as remove + re-add of the same (stable) id — it adds `row` to both `upd$rm` and `upd$add`. The board update path then applies that as remove-then-append, so the link lands last. The `links$mod` channel doesn't help either: `board-server.R` folds it straight into add + rm:

```r
if (length(upd$links$mod)) {
  ...
  upd$links$add <- c(upd$links$add, merged)
  upd$links$rm  <- c(upd$links$rm, names(upd$links$mod))
}
```

Meanwhile `modify_board_links()` already has a position-preserving path — its `mod` arg does `links[names(mod)] <- mod` in place — it's just not on the update path.

### Repro

```r
b <- new_board(
  blocks = c(a = new_dataset_block("BOD"), b = new_dataset_block("BOD"), c = new_rbind_block()),
  links = links(ac = new_link("a", "c", "left"), bc = new_link("b", "c", ""))
)

names(board_links(modify_board_links(b, add = links(ac = new_link("a","c","x")), rm = "ac")))
#> "bc" "ac"   -- ac moved to the end (what the update path does)

names(board_links(modify_board_links(b, mod = links(ac = new_link("a","c","x")))))
#> "ac" "bc"   -- position preserved (the in-place path, unused by the update)
```

### Fix direction

Apply an in-place modification without moving the link. Simplest for existing consumers: treat an id present in both `add` and `rm` as a position-preserving replacement — `modify_board_links()` already has the in-place assignment; the update path just needs to route the overlap through it rather than remove-then-append. Reactive rewiring can still go through the add/rm path in `update_block_links()`; only the board's link order needs preserving.

Both editors — core's manage-links plugin and blockr.dock's edit-board extension (blockr.dock#276) — stage edits as this add∩rm overlap, so the overlap approach fixes both with no consumer change. (If you'd rather make `links$mod` the position-preserving channel instead, both editors would then need to stage edits as `mod`.)

nbenn:
Fixed in `a273570`.

Applied your overlap approach in `modify_board_links()` — an id present in both `add` and `rm` now routes through the in-place replacement (position preserved), with the leftover pure adds/removes handled as before. That fixes the persisted board link order and the `links$mod` fold that lands as the same overlap.

The board order alone wasn't enough, though: the `...args` *container* is rebuilt independently by `variadic_links()` (the resync path added in this PR), and it had the same remove-then-append shape — so the board came out `ac, bc` while the generated code was still `rbind(b, left = a)`. Gave it the same treatment: keep add-ids already present in the board in place, append only genuinely new links. An end-to-end test pins it now — editing the first of two positional `rbind` links keeps `rbind(left = a, b)` instead of flipping to `rbind(b, left = a)`.

Both fixes live in core's shared board machinery, so dock's edit-board extension gets them for free once it stages edits as the same add∩rm overlap — no dock-side change beyond the editor mirror you're already doing.

Full suite green, app tests included.

Issue #55, opened 2025-03-28
Feat: register_block could support local block (without package)


DivadNojnarg:
The current `register_block` does not support when a block does not belong to a package and that's not clear from the roxygen doc:

```r
new_dummy_block <- function(text = "Hello World", ...) {
  new_data_block(
    function(id) {
      moduleServer(id, function(input, output, session) {
        list(
          expr = reactive(quote(text)),
          state = list(text = text)
        )
      })
    },
    function(id) {
      tagList()
    },
    class = "dummy_block",
    ...
  )
}

register_block(
    "new_dummy_block",
    name = c("dummy block"),
    description = c("A block that does nothing"),
    category = c("transform"),
    overwrite = TRUE
  )
```

This can be useful to write vignettes or test blocks without the need of a package.

PR #215, opened 2026-06-05
Resolve the board via a preserve_board loader, not a global slot

### LINKED ISSUE #214: Board loader API bundled with the preserve_board plugin
## Problem

The board reload handoff and the request-phase preload (#153) both stash a board in a process-global slot in core (`serve_obj`), keyed/cached and carrying `meta`, all owned by core. That is the wrong boundary: *how* a board is produced, cached, and carried across a `session$reload()` is blockr.session's concern (persistence/restore), not core's. Core only legitimately knows *when* it needs a board — at the UI build (GET) and the server build (WS connect).

Owning that shared mutable state is also what produced the cross-session contamination fixed in #209: under one shared process the global slot can be read or consumed by the wrong session. #209 makes the slot per-reload-safe; the cleaner fix is for core not to own the slot at all.

## Design

Add a board **loader** as a third component of the `preserve_board` plugin:

```r
preserve_board(server, ui, loader)
# loader: function(request) -> board | NULL   (stateless, request-phase)
```

Core calls the loader at both entry points — `serve_board_ui` at the GET (request carries `req`) and `serve_board_srv` at the WS connect (request carries `session`) — and builds from the returned board, or the `serve()` default when the loader is absent or returns `NULL`.

There is no chicken-and-egg: `serve(board)` supplies the default board, which supplies the plugin set, which supplies the loader; the loader then resolves the actual board. `board_plugins()` is pure, so this works at the GET with no session. Sketch:

```r
loader <- get_plugin("preserve_board", plugins(default))$loader
x      <- if (is.null(loader)) NULL else loader(request)
build(x %||% default, plugins(x %||% default))
```

The loader lives on the plugin, not in a global registry, on purpose. There is exactly one `preserve_board` per board (plugins are one-per-type); it is set declaratively at board construction and replaced only via `custom_plugins()`. So there is no process-global mutable singleton to overwrite — `register_board_preload()` and its overwrite warning go away — and "who may supply the loader" is answered structurally: whoever supplies the board's `preserve_board` plugin.

## What this retires in core

- The `serve_obj` staging store and its accessors (`update_serve_obj()`, `get_serve_obj()`, `finalize_reload()`, `is_reloading()`, the TTL sweep) and the per-reload token / `?__blockr_reload__=` plumbing.
- The `board_refresh` → stage → `session$reload()` producer in `board-server.R`, and `rv$reload_meta`. Reloads are driven by the plugin's own server (it sets the URL and calls `session$reload()` itself); `meta` and anti-loop state move behind the loader, into session.
- `register_board_preload()` — subsumed by the loader, since the preload is just the cold-load case of the same resolution.

## Responsibilities after this

- **core** — lifecycle only: `serve()` / `blockr_app_*`, and "ask the board's `preserve_board` loader to resolve a board for this request, else use the default." Owns no provenance state.
- **`preserve_board` plugin** — `{ui, server, loader}`: UI, in-session logic (including firing its own reloads), and request-phase resolution, plus any private handoff / cache / `meta`.

## Request object

The loader receives a thin uniform wrapper exposing the parsed query (symmetric across both call sites) plus the raw `req`-or-`session` for auth, so the loader signature is stable whether core calls it at the GET or the WS connect. Reconciling the two contexts — including how a backend token is obtained at each — is the loader's (session's) concern; core just passes what it has.

## Relationship to open work

Supersedes the #208 approach (#209 makes the shared slot safe; this removes the slot) and subsumes #153 (the preload hook becomes the loader's cold-load path). The blockr.session counterpart — supplying a loader across the supported backends — is tracked in BristolMyersSquibb/blockr.session#55.

nbenn:
## Summary

- `preserve_board` plugins gain a request-phase **loader** (`function(request) -> board | NULL`) as a third component alongside `server`/`ui`. Core calls it when building the board UI (GET) and server (WS connect) to resolve which board to serve, falling back to the `serve()` default when the loader is absent or returns `NULL`.
- This removes the process-global `serve_obj` slot core used to stage a board across `session$reload()`. Gone: the exported `get_serve_obj()`, the `serve_obj` store and its accessors, the board-server reload producer, and `rv$reload_meta`. `serve()` owns no provenance state — it just asks the board's `preserve_board` loader to resolve a board for the request, else uses the default.
- The default `preserve_board` ships a default loader so core-only UI-triggered restore keeps working: an uploaded JSON is deserialized and staged behind the plugin's own loader, and the plugin's server fires its own `session$reload()` — no core-owned slot, and the whole plugin (loader included) is overridable.
- Supersedes the still-open #154 (`153-double-load`) and #209 (`208-tab-reload-id`), which made the shared slot per-reload-safe rather than removing it; both can be closed.

## Downstream

blockr.session is the loader's real consumer (BristolMyersSquibb/blockr.session#55) and must land in lockstep — it reads `board$reload_meta` and returns a reactive board from its `preserve_board` server, both of which this PR retires; pin session to `core@214-board-loader` until #55 merges. dock and dag are unaffected (and are what core's revdep exercises): `preserve_board` / `restore_board` / `serialize_board` signatures are unchanged and neither reads `reload_meta`.

## Migration

| Before | After |
| --- | --- |
| Core stages a board in `serve_obj`, calls `session$reload()`, next GET/WS reads the slot | The `preserve_board` `loader(request)` resolves the board at GET/WS; `serve()` default when the loader is absent or returns `NULL` |
| The `preserve_board` server returns a reactive board that core consumes to drive a reload | The plugin's server drives its own reload (stages behind its loader, calls `session$reload()`); its return value is no longer consumed |
| `get_serve_obj()` (exported) | removed — the loader owns request-phase resolution |
| reload metadata via `rv$reload_meta` / `board$reload_meta` | moves behind the loader, into the plugin/session |
| `new_plugin()` / `preserve_board()` | gain an optional `loader` argument |

Fixes #214


R/board-plugins.R
```diff
@@ -32,10 +36,10 @@
 #'
 #' @export
 new_plugin <- function(server, ui = NULL, validator = abort_not_null,
-                       class = character()) {
+                       class = character(), loader = NULL) {
```
nbenn: Hard-coding a `loader` arg for all plugins is not an option. Let's allow passing `...` instead.

R/board-plugins.R
```diff
@@ -32,10 +36,10 @@
...
 
   res <- structure(
-    list(server = server, ui = ui),
+    list(server = server, ui = ui, loader = loader),
```
nbenn: Let's absorb per-plugins args as attributes and not place them into the main object.

nbenn:
Cross-referencing the loader API against our earlier design discussion. The pivot is sound: for the session/production path the loader resolves the board from the request, so there is no shared slot and both the #208 cross-session contamination and the reload loop are gone by construction, and the `board_request` abstraction (`$query` normalized across GET/WS, `$session` NULL at the GET) cleanly handles the UI-vs-server asymmetry.

(Posting as one top-level comment rather than inline: there is an open pending review on this PR, and GitHub allows only one pending review per account, which blocks inline review comments from me.)

Two issues to address, plus a doc note: the loader receives `request$session == NULL` at the GET — worth stating in the loader contract so third-party loaders are written to tolerate it.

---

**1. `R/utils-serve.R`, `resolve_board()` (line 247) — the loader return is unvalidated.**

`coal(board, default, fail_all = FALSE)` passes whatever the loader returns straight through, and `check_ser_deser_val` — which validated the return and called `validate_board()` — was removed with nothing replacing it. A loader that returns a non-`NULL`, non-board (a bug in a third-party loader) becomes the served board, which then blows up in `blockr_app_ui()` / `board_server()`. This is exactly the `validate_preload_result` hardening that went into #154, dropped by the pivot. Suggest validating here: accept `NULL` or `is_board()`, `blockr_abort()` with a clear class otherwise, and probably run `validate_board()` on the resolved board.

**2. `R/plugin-serdes.R`, default `reload_handoff` (line 96) — single slot, preview-grade, undocumented.**

This default handoff is a single process-global slot, so two concurrent upload-restores in a core-only app contaminate each other exactly as #208 did (and it regresses the per-tab fix #209 had for the upload case): a fresh session that connects while a board is staged reads it, and the booting server then clears it, stranding the original. That is a defensible trade for a *default* — session apps replace the loader, and a single slot can't leak unboundedly the way per-token did — but it's currently undocumented. Worth one sentence in NEWS and the `preserve_board` docs: the default save/restore handoff is single-tab / preview-grade; for multi-user, supply a `loader`.

nbenn:
Thanks — all three addressed in 420119c.

**1. Loader-return validation.** `resolve_board()` now returns the `serve()` default on `NULL`, `blockr_abort()`s with class `invalid_board_loader` for a non-`NULL` non-board, and runs `validate_board()` on the resolved board — restoring the `validate_preload_result` hardening the pivot had dropped. Added a regression test (`resolve_board rejects a non-board loader result`).

**2. Preview-grade default handoff.** Documented in NEWS and the `preserve_board` topic: the default save/restore handoff is a single process-global slot (single-tab / preview-grade), so concurrent restores in one process can clobber one another, and multi-user deployments should supply their own keyed `loader` (as blockr.session will via #55).

**3. GET-phase `session`.** The loader contract now states `request$session` is `NULL` at the GET (no session yet), so a loader must tolerate its absence — in both the `preserve_board` description and the `@param request` for `preserve_board_loader`.

R/board-plugins.R
```diff
@@ -113,6 +118,10 @@ plugin_server <- function(x) x[["server"]]
 #' @export
 plugin_ui <- function(x) x[["ui"]]
 
+#' @rdname new_plugin
+#' @export
+plugin_loader <- function(x) x[["loader"]]
```
nbenn: No need for this. At least not at the "plugin"-level. I'd be fine with such a utility at the `preserve_board`-level.

R/plugin-serdes.R
```diff
@@ -65,11 +87,48 @@ preserve_board_server <- function(id, board, ...) {
         }
       )
 
-      res
+      observeEvent(
+        res(),
+        {
+          val <- res()
+
+          stage_reload_handoff(if (is_board(val)) val else val$board)
```
nbenn: The current implementation is off: we expose an argument `loader`, but whatever is passed there feeds into hard-coded internals here. This needs proper encapsulation.

R/utils-serve.R
```diff
@@ -240,39 +238,48 @@ serve_board_srv <- function(id, plugins, options, ...) {
...
 
-  rm(list = id, envir = serve_obj, inherits = FALSE)
-  invisible(obj$meta)
+  validate_board(board)
 }
 
-#' @rdname serve
-#' @export
-get_serve_obj <- function(id = NULL) {
+new_board_request <- function(query, request, session) {
```
nbenn: This seems a bit over-engineered to me. Do we really need this object?

nbenn (commented):
Comments inline.

nbenn:
Reshaped per the discussion — pushed in b181117.

**Plugin structure (your inline points 1–3).** `new_plugin()` no longer hard-codes a `loader` field: it takes `...` for plugin-specific components, attached as attributes; the list stays `{server, ui}`. `preserve_board()` passes its `loader` through there, and the generic `plugin_loader()` is gone — `preserve_board` reads it via an internal `board_loader()` accessor.

**The channel is the request (points 4–5).** No process-global handoff — it couldn't match the right before/after session across processes. The board crosses `session$reload()` through the request: the default save/restore mints a one-time token, stages the uploaded board under it, writes the token into the URL, and reloads; the loader reads that token from `request$query` and resolves per-tab, and the token is stripped + the entry consumed once the new session initializes (so a clean reload, no token, falls back to the default — no 2× load). A downstream `preserve_board` (blockr.session) resolves instead from the URL's `board_name`/`version` against its backend, which is multi-process-safe. Both subscribe to the same channel — the request — differing only in key (token vs. natural id) and store (in-process vs. backend). As agreed, mixing different `preserve_board` plugins across one reload is unsupported.

The `board_request` object is gone: the loader receives a plain `list(query, request, session)` (`session` is `NULL` at the GET).

Full suite green (1280), lint clean, docs regenerated.

nbenn:
Done — the `board_loader` object, shuttled by `serve()`, is in 83cf571.

`board_loader(resolve, stage)` bundles `resolve(request)` and `stage(board, session)` over a shared store. `serve()` pulls the served board's loader off its `preserve_board` once and uses that single instance for both the request-phase resolution (in `serve_board_ui` / `serve_board_srv`) and the in-session staging (threaded into `board_server`, where the `preserve_board` server calls `loader$stage`).

That's what dissolves the stable-plugin constraint we were circling: since the *same* loader drives resolve and stage, the board crossing a reload is independent of which `preserve_board` the resolved board itself carries — the resolved board keeps its own UI/server, only the handoff is pinned to the app's loader. And the default loader keeps its token store in its own closure, which rides in the app object, so it survives the reload with no process global at all and is scoped per-`serve()` (two apps in one process don't share a slot). A downstream loader (blockr.session) supplies its own `resolve` = query→backend / `stage` = write-URL, which is multi-process-safe and never touches the in-process store.

`board_loader()` / `is_board_loader()` are exported. One contract note: the `preserve_board` server now receives the app loader, so a custom server takes it via `...` or a `loader` argument (the test fixtures and the default server do).

Full suite green bar one local chromote flake on a `serve()` AppDriver test (the app starts fine — `serve()`, UI build and `board_server` all verified for that board without the browser; it's a 4s-timeout under local load); CI runs those clean. Lint clean, docs regenerated.

nbenn:
Done — moved in fe11604, and yes, this is much closer to the original shape (with the loader as the genuine improvement).

**Loader is now a `board_loader` option, not a plugin field.** `serve()` reads `blockr_option("board_loader", <in-process default>)` once. `new_plugin()` and `preserve_board()` are back to `{server, ui}` — the plugin no longer carries a loader, and the `...`/attribute machinery is gone. This is the root fix for what you spotted: the plugin re-creates on every build, so it was never where the loader survived a reload — the option is, and "an option I set configures the app for its life" is exactly the right user model.

**Reload is core's again.** The `preserve_board` server is a pure producer returning the board to restore (`check_ser_deser_val` is back). `board_server` observes it, calls `loader$stage(board, session)` to persist and get back the URL query params, and **core** writes them into the URL and runs `session$reload()`. The loader never reloads; the teardown-sensitive path stays in core.

**The two-incompatible-plugins scenario dissolves entirely.** There's one app-level loader, the plugin carries none, and the server never touches the loader — so a board "using a different plugin" has nothing to be incompatible with.

**Validation, as asked.** `board_loader(resolve, stage)` rejects (with classed errors) a `resolve` that isn't a function of `request` or a `stage` that isn't a function of `board`/`session` (`...` allowed); `resolve`'s return is validated in `resolve_board` (`NULL`/board, else `invalid_board_loader`).

Full suite green, lint clean, docs regenerated. session sets the loader via `options(blockr.board_loader = …)`.

R/board-server.R
```diff
@@ -263,55 +263,15 @@ board_server.board <- function(id, x, plugins = board_plugins(x),
...
-            board <- val
-            meta <- NULL
-          } else {
-            board <- val$board
-            meta <- val$meta
+        observeEvent(
+          board_refresh(),
+          {
+            val <- board_refresh()
+            reload_board(loader, if (is_board(val)) val else val$board, session)
```
nbenn: Can we not be stricter here? The whole `val$meta` branch might not be needed anymore. This was a workaround for limited customizability that should go away with the new approach afaiu.

R/plugin-serdes.R
```diff
@@ -25,6 +41,48 @@ preserve_board <- function(server = preserve_board_server,
...
+
+  structure(
+    list(resolve = resolve, stage = stage),
+    class = "board_loader"
+  )
+}
+
+has_formals <- function(f, args) {
+  nms <- names(formals(f))
+  "..." %in% nms || all(args %in% nms)
```
nbenn: Should we not disallow `...`?

R/utils-serve.R
```diff
@@ -240,39 +245,73 @@ serve_board_srv <- function(id, plugins, options, ...) {
   }
 }
 
-serve_obj <- new.env()
+resolve_board <- function(default, loader, request) {
+
+  board <- if (is.null(loader)) NULL else loader$resolve(request)
```
nbenn: Do we really need to handle the `is.null(loader)` case? What does this mean if the preserve_board triggers a reload? You always end up with the starting board? This hardly seems like good UX.

R/utils-serve.R
```diff
@@ -240,39 +245,73 @@ serve_board_srv <- function(id, plugins, options, ...) {
...
+      "A `preserve_board` loader must return `NULL` or a `board`, but got ",
+      "{class(board)} instead.",
+      class = "invalid_board_loader"
+    )
+  }
+
+  validate_board(board)
+}
+
+board_request_get <- function(request) {
```
nbenn: I don't like the indirection here. Why do we need to create such an arbitrarily shaped "request" object? Why don't we call `loader$resolve()` with proper args? What are the actually needed args anyways? Do `query` and `session` not suffice? What's the distinction between `query` and `request`?

R/utils-serve.R
```diff
@@ -103,12 +102,13 @@ serve.block <- function(x, id = "block", ..., data = list()) {
 serve.board <- function(x, id = rand_names(), plugins = blockr_app_plugins,
                         options = blockr_app_options, ...) {
 
-
   stopifnot(is_string(id), is.function(plugins), is.function(options))
 
+  loader <- blockr_option("board_loader", preserve_board_loader())
```
nbenn: If you start threading through the loader from here, why not make it an argument to `serve()`, with `preserve_board_loader()` as default? It would have to be exported and we for sure need a better name. Something that describes how it works.

R/plugin-serdes.R
```diff
@@ -25,6 +41,48 @@ preserve_board <- function(server = preserve_board_server,
...
+#' `function(board, session)` which persists `board` and returns the URL query
+#' parameters referencing it (core writes them and reloads). They share private
+#' state, so a `board_loader` is built as a unit, not supplied as two loose
+#' functions.
+#'
+#' @rdname preserve_board
+#' @export
+board_loader <- function(resolve, stage) {
+
+  if (!is.function(resolve) || !has_formals(resolve, "request")) {
```
nbenn: Please out-factor this into a validator. A pattern in use throughout the code-base here. Then construct the object, validate and return.

R/utils-serve.R
```diff
@@ -103,12 +102,13 @@ serve.block <- function(x, id = "block", ..., data = list()) {
...
+
```
nbenn, line 111: Here would a a great place to *validate* the passed `loader` object. See also https://github.com/BristolMyersSquibb/blockr.core/pull/215/changes#r3467174864.

R/board-server.R
```diff
@@ -37,16 +37,17 @@ board_server <- function(id, x, ...) {
...
 #' plugins)
+#' @param loader The app's [board_loader()] (supplied by [serve()]), through
+#' which the `preserve_board` server stages a board for the next reload
 #' @rdname board_server
 #' @export
 board_server.board <- function(id, x, plugins = board_plugins(x),
                                options = board_options(x),
                                callbacks = list(),
                                callback_location = c("end", "start"),
+                               loader = NULL,
```
nbenn: Given that this is part of the public API, please *validate* the passed loader here. See also https://github.com/BristolMyersSquibb/blockr.core/pull/215/changes#r3467174864. Also, what is it with the `NULL` default? Again, you then perhaps have some `preserve_board` UI that is non-functional. Bad UX. Could we get away with not "contaminating" this entire scope with the loader surface? Something like have `board_server()` return `board_refresh` and handle the reload as part of `serve_board_srv`? I'm not sure this is a good idea, but might be something worth exploring.

nbenn:
Thanks — all eight addressed in c9029f9. In order:

**`utils-serve.R:107` — loader as a `serve()` argument + rename.** Done: `serve(board, …, loader = local_loader())`. The `blockr_option` is gone; the loader is a plain argument, validated in `serve()`. I renamed `preserve_board_loader()` → `local_loader()` — it resolves and stages in-process against a per-instance store, "local" as opposed to a shared backend. Flagging the name for you: if you'd prefer something that names the token-handoff mechanism more explicitly (`reload_loader`, `inproc_loader`, …), say the word.

**`plugin-serdes.R:55` + `utils-serve.R:108` — out-factor a validator, validate at the entry point.** New `validate_board_loader()` does construct-validate-return; `board_loader()` calls it, and so does `serve()` on the passed loader.

**`board-server.R:48` — validate the loader, the `NULL` default, and the explore.** I took the exploration, and it's cleaner. `board_server()` no longer takes a loader and no longer reloads — it just exposes the board to restore as a `board_refresh` reactive in its return. `serve_board_srv()` observes that and calls `reload_board()`. So the loader surface is confined to `serve()` (one place to validate), the bad-UX `NULL` default disappears, and the reload is still core's. If you'd rather keep it in `board_server()`, it's an easy revert.

**`plugin-serdes.R:77` — disallow `...`.** `has_formals()` now requires the named formals; a bare `…` no longer satisfies it.

**`utils-serve.R:267` — the arbitrary request object.** Dropped. `resolve(query, session)` is called with proper args; `board_request_get/ws` are gone. To your questions: a loader needs only the parsed `query` (the board reference / token) and the `session` (`NULL` at the GET, set at the WS connect — and at the WS it carries `session$request`, so header-based auth is reachable there). The raw Rook request added nothing query+session didn't, which is exactly the `query`-vs-`request` confusion — there's now just `query`.

**`utils-serve.R:250` — the `is.null(loader)` case.** Removed. You're right it was bad UX: a staged restore would silently resolve back to the starting board. The loader is always present now (the `serve()` default), so `resolve_board()` always calls `loader$resolve()`.

**`board-server.R:272` — the `val$meta` branch.** Gone, and I pulled the rest of the dead `meta` plumbing with it: `restore_board()` loses its `meta` argument, and a `preserve_board` server must now return a board (not `list(board, meta)`). One heads-up for a dock follow-up: `restore_board.dock_board()` still carries a `meta` parameter, but core never populates it (the producer's `read_plugin_args` has no `meta`), so it always takes its `is.null(meta)` branch and returns a plain board — the revdep stays green and the param is simply vestigial there now.

Full suite green under `R CMD check` (the faithful runner this time, not just `load_all`), lint clean, docs regenerated.

R/plugin-serdes.R
```diff
@@ -25,6 +41,67 @@ preserve_board <- function(server = preserve_board_server,
...
+      "A `board_loader` `stage` must be a function of `board` and `session`.",
+      class = "board_loader_stage_invalid"
+    )
+  }
+
+  x
+}
+
+has_formals <- function(f, args) {
+  all(args %in% names(formals(f)))
```
nbenn: We could be more strict here, no? Everything short of `identical()` is bad news.

R/plugin-serdes.R
```diff
@@ -70,6 +147,59 @@ preserve_board_server <- function(id, board, ...) {
...
+
+      strip_reload_token(session)
+    }
+
+    board
+  }
+
+  stage <- function(board, session) {
+
+    token <- rand_names()
```
nbenn: To have a guarantee that the token does not already exist, you could pass `ls(...)` output to `rand_names()`.

R/utils-serve.R
```diff
@@ -218,61 +222,88 @@ serve_board_ui <- function(id, plugins, options, ...) {
...
     invisible()
   }
 }
 
-serve_obj <- new.env()
+resolve_board <- function(default, loader, query, session) {
+
+  board <- loader$resolve(query, session)
+
+  if (is.null(board)) {
```
nbenn: How would this happen? Is this not reason to error outright? I wonder a bit whether we need this function at all: `validate_board(loader$resolve(query, session))` is all it does, no?

R/utils-serve.R
```diff
@@ -218,61 +222,88 @@ serve_board_ui <- function(id, plugins, options, ...) {
...
 
-finalize_reload <- function(id = "reload") {
+reload_board <- function(loader, board, session) {
 
-  obj <- get0(id, envir = serve_obj, inherits = FALSE)
+  params <- loader$stage(board, session)
 
-  if (is.null(obj)) {
-    return(invisible(NULL))
+  if (length(params)) {
```
nbenn: This is an open question, not a request: Would it not make sense to move this logic into the staging function? Having to update query params is needed for the default implementation, but not necessarily for others. And if it is, they have `session` available to modify.

R/utils-serve.R
```diff
@@ -218,61 +222,88 @@ serve_board_ui <- function(id, plugins, options, ...) {
...
-  exists(id, envir = serve_obj, inherits = FALSE)
+session_query <- function(session) {
+  parseQueryString(coal(isolate(session$clientData$url_search), ""))
 }
 
-finalize_reload <- function(id = "reload") {
+reload_board <- function(loader, board, session) {
 
-  obj <- get0(id, envir = serve_obj, inherits = FALSE)
+  params <- loader$stage(board, session)
```
nbenn: The `stage` component might be optional, no? We don't necessarily need this for other loader implementations. If this function is convenient for unit tests, keep it. If not, it feels like a bit overkill if we decide to fold the query update mechanism nto the default loader (https://github.com/BristolMyersSquibb/blockr.core/pull/215/changes#r3467709073).

R/plugin-serdes.R
```diff
@@ -25,6 +41,67 @@ preserve_board <- function(server = preserve_board_server,
...
+#' `function(query, session)` returning the `board` to build or `NULL` (the
+#' parsed URL `query`, and the `session` -- `NULL` at the GET, set at the WS
+#' connect); `stage` is `function(board, session)` which persists `board` and
+#' returns the URL query parameters referencing it (core writes them and
+#' reloads). They share private state, so a `board_loader` is built as a unit,
+#' not supplied as two loose functions.
+#'
+#' @rdname preserve_board
+#' @export
+board_loader <- function(resolve, stage) {
```
nbenn: If we decide the make the `stage` component optional, we'd need a `NULL` default value.

R/plugin-serdes.R
nbenn: Done in 5242302 — `has_formals` is now `formals_match`, using `identical(names(formals(f)), args)`, so a loader's `resolve`/`stage` formals must match the contract exactly; extra arguments or a bare `...` are now rejected. (The loader machinery also moved to its own `board-loader.R` in 44583d8, so this lives there now.)

R/plugin-serdes.R
nbenn: Done in 5242302 — `token <- rand_names(ls(envir = store))`, so the freshly drawn token is guaranteed not to collide with a board already staged in this loader's store.

R/utils-serve.R
nbenn: It is `validate_board(resolve())` plus two things that earn it its keep. First, the `NULL` → `serve()`-default fallback: a loader returning `NULL` means "nothing staged for this request, build the default", and `validate_board(NULL)` would error rather than fall back. Second, a readable contract error when a loader returns a non-board — I tried leaning on `validate_board()` for that, but since it dispatches on class, a non-board only triggers R's "no applicable method for 'validate_board'", so I kept the explicit `is_board()` guard with a clear message. And to your first question: a non-board can only come from a buggy loader, so erroring outright is exactly right (which it does). Happy to inline it at the two call sites if you'd still rather drop the helper. It now lives in `board-loader.R`.

R/utils-serve.R
nbenn: I weighed folding it in and landed on keeping the write in `reload_board`, for two reasons. `stage` already receives `session`, so a loader that needs to do something other than set query params (hit a backend, set a cookie) can already do it there — and a loader that needs no URL params just returns an empty list, which `reload_board` skips; so "others may not need this" is handled without making every loader own the write. And keeping `stage` a pure `(board, session) -> params` function keeps the default loader unit-testable: the test gets the token back from `stage` and drives `resolve` with it. If `stage` wrote to the URL and returned nothing, the token would be private and that round-trip would only be reachable through a live reload. The `session$reload()` itself stays in core either way. It is genuinely a judgment call, though — if you'd still prefer `stage` to own the write, I'll move it.

R/utils-serve.R
nbenn: `stage` is now optional (the `board_loader(resolve, stage = NULL)` change), and `reload_board` guards on it before staging. I kept `reload_board` as a named helper rather than inlining it — it is exactly the unit-test seam you mention (mock the loader, assert it stages then reloads), and since I didn't fold the query write into `stage` (the adjacent thread) it still carries the generic write-then-reload. If we do decide to fold, it collapses to a one-liner and I'll inline it then.

R/plugin-serdes.R
nbenn: Done in 5242302 — `board_loader(resolve, stage = NULL)`, and `validate_board_loader` now accepts a `NULL` stage (otherwise it must be a `function(board, session)`).

nbenn:
From the blockr.session loader (`board_loader`'s real consumer, BristolMyersSquibb/blockr.session#56): `resolve` is missing two inputs only core can supply — one of them with no downstream workaround.

**The raw request at the GET — the must-have.** `resolve(query, session)` never receives the HTTP request, and at the GET `session` is `NULL`, so request headers are unreachable there. That is exactly what credential-layer auth needs (a Connect viewer token / vault JWT rides in request headers) — the access model we'd rather loaders use than app-logic gating — and only core can provide it: `serve_board_ui` holds the request but throws it away after pulling `QUERY_STRING` ([utils-serve.R#L212-L213](https://github.com/BristolMyersSquibb/blockr.core/blob/524230231c9400c97ddc4781dbe2dcf6b2ee7b99/R/utils-serve.R#L212-L213)). At the WS a loader can reach `session$request`; the GET has nothing. So `resolve` should receive the raw `request` at both phases (the UI function's `request`, and `session$request` at the WS).

**The served board — nice-to-have.** A loader often needs the board `serve()` was given, e.g. to derive its own default (`clear_board(initial_board)` for a blank "New" / cold load). Core has it as `default` in [`resolve_board`](https://github.com/BristolMyersSquibb/blockr.core/blob/524230231c9400c97ddc4781dbe2dcf6b2ee7b99/R/board-loader.R#L141) but only ever uses it as the NULL fallback, so the loader can't see it. blockr.session currently works around this by capturing the board and passing it twice (`serve(board, loader = f(board))`); handing it to `resolve` removes the redundancy.

**Drop the pre-parsed `query`.** With the raw request in hand the query is derivable, and carrying `query` and `request` side by side is more confusing than extracting it. One caveat for whoever implements: the extraction is phase-split — `request$QUERY_STRING` at the GET, but `session$clientData$url_search` at the WS (the websocket request does not carry the page query, which is why [`session_query`](https://github.com/BristolMyersSquibb/blockr.core/blob/524230231c9400c97ddc4781dbe2dcf6b2ee7b99/R/board-loader.R#L159) reads `clientData`). Exporting a generalized `session_query(request, session)` helper would spare every loader that footgun, but it is optional — we are happy to extract.

Proposed shape, as plain args (no bundled request object):

```r
resolve <- function(request, session, initial_board) { ... }
```

`request` raw at both phases, `session` `NULL` at the GET, `initial_board` the served board (name negotiable — `default` / `x` if that is more consistent internally). `validate_board_loader`'s `formals_match` ([board-loader.R#L36](https://github.com/BristolMyersSquibb/blockr.core/blob/524230231c9400c97ddc4781dbe2dcf6b2ee7b99/R/board-loader.R#L36)) updates to match; `stage` is unaffected.

Happy to adopt it on the blockr.session side once it lands — the loader goes back to a plain resolve-only `board_loader` with no board capture.

R/utils-serve.R
nbenn: You were right, and I was wrong to resist this. Folded the `updateQueryString` into `local_loader`'s `stage` — it now persists the board *and* writes the URL, so the query-param logic no longer lives in core at all. My testability objection doesn't hold up: the token round-trip is still unit-tested, the test just captures what `stage` writes through a fake `session$updateQueryString`. The only piece kept in core is the bare `session$reload()` (your hard line — the reload is the guarantee), now inlined in `serve_board_srv`.

R/utils-serve.R
nbenn: Gone. With the URL write folded into `stage`, the stage-then-reload collapses to two lines inlined in `serve_board_srv` — `if (not_null(loader$stage)) loader$stage(board, session)` then `session$reload()`. No `reload_board`, and `session$reload()` stays the bare, unconditional core call so the reload guarantee holds regardless of what a loader's `stage` does (or whether it has one).

nbenn:
Implemented — `resolve(request, session, default)`, in 531742b. And following up on your `session_query` question below: you're right that it wasn't doing enough to justify exporting, so it's gone from the API.

**Raw request at both phases.** `serve_board_ui` hands its `request` to `resolve` (GET, `session = NULL`); `serve_board_srv` passes `session$request` (WS). So GET-phase header auth — the must-have — works.

**Served board as `default`.** `resolve` now receives the board `serve()` was given, named `default` to match what core already calls it everywhere (per your nudge), rather than a second name. blockr.session can drop the capture-and-pass-twice and derive its own result (`clear_board(default)`) directly.

**The loader mechanism is fully encapsulated now.** `session_query` is un-exported, and all the URL machinery (query read, query-string build, token write/strip) moved into `local_loader` as private closures — none of it is any other loader's concern. Another loader extracts its own query; the only thing it needs to know is the phase split (`request$QUERY_STRING` at the GET, `session$clientData$url_search` at the WS), which is now documented as guidance instead of shipped as a helper. `reload_board` is gone too — the URL write is folded into `stage`, and `session$reload()` stays the bare core call in `serve_board_srv` (the one piece that has to remain a guarantee). Details on the two threads.

`validate_board_loader`'s `formals_match` requires `c("request", "session", "default")`; `stage` is unchanged.

PR #225, opened 2026-06-16
Point CI workflows at BristolMyersSquibb/blockr.ci

### LINKED ISSUE #224: Update blockr.ci references after its transfer to the BristolMyersSquibb org
`blockr.ci` was transferred from the `cynkra` org to `BristolMyersSquibb`, so this repository has workflows that still reference it by the old path. GitHub redirects `cynkra/blockr.ci` to `BristolMyersSquibb/blockr.ci`, so CI keeps working for now — but the references should be updated explicitly: the redirect silently breaks if the `cynkra/blockr.ci` name is ever reclaimed, and the explicit path reads clearer.

Swap the `cynkra` owner for `BristolMyersSquibb` in the `uses:` references (keep the path and `@ref` unchanged) in:

- [ ] `.github/workflows/ci.yaml`
- [ ] `.github/workflows/pkgdown.yaml`

The reusable-workflow and job names are unchanged, so no required-status-check or ruleset updates are needed.

nbenn:
## Summary

- `blockr.ci` moved from the `cynkra` org to `BristolMyersSquibb`. CI only kept working via GitHub's owner redirect, which reads less clearly and silently breaks if the `cynkra` name is ever reclaimed.
- Repoint every reusable-workflow `uses:` reference to `BristolMyersSquibb/blockr.ci`, keeping the path and `@main` ref unchanged: `ci.yaml` (both the `ci` and `revdep` jobs) and `pkgdown.yaml`.
- Reusable-workflow and job names are unchanged, so no required-status-check or ruleset updates are needed.

Fixes #224

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,7 +7,7 @@ name: ci
 
 jobs:
   ci:
-    uses: cynkra/blockr.ci/.github/workflows/ci.yaml@main
+    uses: BristolMyersSquibb/blockr.ci/.github/workflows/ci.yaml@main
     with:
       lintr-exclusions: |
         vignettes/create-block.qmd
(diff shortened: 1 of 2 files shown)


nbenn:
Superseded by #226, which merged the byte-identical change to main (cc6c5a71). No diff remains against main, so closing this as a duplicate.

PR #233, opened 2026-06-22
Add structured block argument specs and registry metadata fields

### LINKED ISSUE #121: New fields in registry?
How to add new fields to the registry? I would like to have an (optional) `link` field to point to the help page in the documentation. This would allow me to include it in an updated help popover:

Image

Similarly, in addition to `name` and `description`, a new optional field `details` would allow us to make the popover more detailed, giving the user all the details and perhaps make the link to the help page superfluous for most block.

Alternatively, could we make this more flexible and allow for arbitrary new fields. For example, for the AI blocks, some block specific 'skill' instructions may be useful, and it could make sense to give block authors the possibility to define new fields.

nbenn:
## Summary

- The construction metadata the LLM tooling depends on -- per-argument worked examples and block-level guidance -- was carried as undocumented `examples` / `prompt` attributes on `arguments`, which core never declared or validated and which are silently dropped when a consumer round-trips `arguments` through a data frame. A third-party block author can't discover the convention, and that lost guidance is why blockr.assistant can't configure blocks.
- Replace the attribute side-channel with a core-owned, validated specification: `block_args()` / `block_arg()` carry a per-argument `description`, a single worked `example`, and an optional opaque `type`. `register_block()` gains block-level `details`, `link`, `guidance`, `examples`, and `keywords`.
- Hard-validated at registration (which runs at package load, so violations fail app startup): every argument name in a user-supplied spec must be a real constructor formal, and every worked example must actually construct.
- Backward compatible. Bare named character vectors and `character()` still work; the legacy `examples` / `prompt` attributes are absorbed into the spec with a one-time deprecation warning; and `registry_metadata(.., "arguments")` keeps projecting the legacy named-vector shape (with `examples` / `prompt` attributes reconstructed), so unmigrated producers and consumers are unaffected. `block_arg_specs()` and `block_examples()` expose the structured form for new consumers.

The finalized design, and the per-argument-vs-block-level reasoning behind it, is in [this issue comment](https://github.com/BristolMyersSquibb/blockr.core/issues/121#issuecomment-4767120119). Producer migration (blockr.dplyr / ggplot / io) and consumer migration (blockr.ai, blockr.assistant) are tracked separately in blockr.ai#86 and blockr.assistant#54.

Fixes #121


R/block-args.R
```diff
@@ -0,0 +1,376 @@
...
+  UseMethod("block_examples")
+}
+
+#' @export
+block_examples.block_registry_entry <- function(x, ...) {
+  block_examples_list(attr(x, "arguments"), attr(x, "examples"))
+}
+
+#' @export
+block_examples.character <- function(x, ...) {
```
nbenn: instead of duplicating `.block`/`.character` methods here over all getters, why not have a single `block_metadata()` generic that does this resolving and then do something like `block_examples.default <- function(x, ...) block_examples(block_metadata(x, ...))`?

nbenn (commented):
We added getters for the new registry fields only. What about the existing?

R/block-args.R
```diff
@@ -0,0 +1,376 @@
...
+}
+
+block_arg_field <- function(x, field) {
+  stopifnot(inherits(x, "block_arg"))
+  x[[field]]
+}
+
+#' @rdname register_block
+#' @export
+arg_description <- function(x) {
```
nbenn: This should be `block_arg_description()` I think. Also, Why not make that generic too? we get for free resolving this from e.g. a string.

R/block-args.R
```diff
@@ -0,0 +1,339 @@
...
+  x
+}
+
+#' @export
+as_block_arg.character <- function(x, ...) {
+  new_block_arg(description = x)
+}
+
+#' @export
+as_block_arg.default <- function(x, ...) {
```
nbenn: A .default method that throws is pointless. We get that for free already.

R/block-class.R
```diff
@@ -735,35 +735,102 @@ has_external_ctrl <- function(x) {
...
+block_metadata_record <- function(x, ...) {
+  UseMethod("block_metadata_record")
+}
+
+#' @export
+block_metadata_record.block <- function(x, ...) {
+
+  record <- attr(x, "block_metadata")
+
+  if (is.list(record)) record else list()
```
nbenn: Why an empty list? Let's error if we don't get a proper result here if anything!

R/block-class.R
```diff
@@ -735,35 +735,102 @@ has_external_ctrl <- function(x) {
...
+block_metadata_record.character <- function(x, ...) {
+
+  record <- block_metadata_record(get_registry_entry(x))
+  record[["id"]] <- x
+
+  record
+}
+
+#' @export
+block_metadata_record.default <- function(x, ...) {
```
nbenn: Drop it.

R/block-args.R
```diff
@@ -0,0 +1,339 @@
...
+#' @rdname register_block
+#' @export
+block_guidance <- function(x, ...) {
+  coal(block_metadata_record(x, ...)[["guidance"]], character(),
+       fail_all = FALSE)
+}
+
+#' @rdname register_block
+#' @export
+block_keywords <- function(x, ...) {
```
nbenn: This belongs in another file, no? It has nothing to do with block_args.

nbenn (commented):
We have the convenient df-returning wrapper `registry_metadata()` and per field accessors like `block_guidance()`, `block_keywords()`, etc. What is missing for completeness here? Should we add `block_description()`, `block_icon` etc.? Where do we draw the line? Or do we remove all field accessory that are already covered by `registry_metadata()`?

christophsax:
# PR #233 review — `new_block_arg(type=)` should be a harness-agnostic content-structure spec, not an ellmer type

Overall this is a great direction — promoting block construction metadata to a first-class, validated registry form is exactly right, and the spike (registering blockr.dplyr's `select`/`filter` against this branch) shows the structured args, `guidance`, `keywords` and example-construct validation all working, with the legacy attribute form still absorbed. One thing I'd push back on before merge: the **`type`** field.

## The concern

The field is documented as *"Optional machine-readable type (e.g. an `ellmer::type_*`), **stored opaquely** and not interpreted by blockr.core"*. Putting a live `ellmer` type object in the registry is, I think, the wrong call, for three reasons:

**a) It forces an unwanted dependency into block packages.** To declare `type = ellmer::type_enum(...)`, blockr.dplyr (a foundational data-transform package with no business knowing about LLMs) would have to add `ellmer` to its Imports. Every typed block package inherits that. ellmer is a heavy dependency to pull into the base data layer purely to annotate arguments.

**b) It's the wrong layer.** The registry's job is to describe **what an argument contains** — its content structure. It should not encode the type system of whichever *harness* happens to consume it. ellmer is one consumer (blockr.ai); blockr.mcp, a raw OpenAI/Anthropic tool schema, or anything else are others. Baking ellmer in means the block author is reaching across three layers to speak the harness's dialect. The block should state the shape; the harness should derive *its* types from that shape.

**c) It forfeits validation.** The reason `type` has to be *stored opaquely* is precisely that it's a foreign ellmer object core can't understand — so core can only hold it and hand it back, never check it. That opacity is a symptom of the ellmer choice, not a feature: a wrong/stale type passes registration silently, and there's no way to confirm the type even agrees with the worked example. A type vocabulary that core *owns* is no longer opaque — core can validate it at registration (like it already validates examples) and cross-check the example against it.

## Suggested direction

Keep the `type` slot, but instead of an opaque foreign object, make its content a small, dependency-free, declarative descriptor of structure and meaning that **core owns and validates**, via core-provided constructors:

```r
# instead of ellmer::type_enum(c("&","|"))
arg_enum(c("&", "|"))

# filter$conditions — polymorphic array of records
arg_array(arg_record(
  type   = arg_enum(c("values","numeric","expr")),
  column = arg_column(required = FALSE),       # semantic: a data column
  values = arg_array(arg_literal(), required = FALSE),
  op     = arg_enum(c(">",">=","<","<="), required = FALSE),
  expr   = arg_expression(required = FALSE)     # semantic: R code
))

# select$columns
arg_array(arg_column())
```

Vocabulary (`kind`): `enum / array / record / column / expression / literal / string / number / boolean`, plus a `required` flag.

Three payoffs beyond removing the dependency:

- **It's *more* expressive than ellmer types, not less.** `column` and `expression` carry semantic meaning ellmer has no concept of (both would just be `type_string`). A harness can use that to constrain values to the data's actual columns, or to validate/format an R expression — not possible from a bare `type_string`.
- **It's harness-agnostic.** blockr.ai maps `enum→type_enum`, `record→type_object`, `column/expression/literal→type_string(+description)`; MCP or a raw tool schema map the same spec their own way. One source of truth, no consumer privileged.
- **Core can validate it** (because it's no longer opaque). At registration, check the `kind` is known and the spec is well-formed — and cross-check the example *against* the declared type. That closes a gap in the current example check: it's construct-only, so for permissive flat-args constructors a malformed-but-constructible example — e.g. `conditions = "not-a-list"` — passes silently today (the guarantee is "doesn't error the constructor", not "is valid"). With a typed `conditions` (array-of-records), core could reject it.

So: the block package describes the content structure once, dependency-free; core validates it; each harness binds it to its own type system. Happy to prototype the `arg_*()` constructors + a blockr.ai translator if useful.

nbenn:
Thanks @christophsax — this is the right critique and I'm convinced on the core of it. The opaque `ellmer::type_*` is going, for exactly the three reasons you give: it forces an LLM dependency into foundational block packages, it encodes one harness's dialect at the wrong layer, and "stored opaquely" is precisely what forfeits validation. And I'm taking your suggested direction: core owns a small, dependency-free, validated descriptor of an argument's content structure, and each harness binds that to its own types.

One change to the direction, and one thing I want to hold off on.

The change: rather than a vocabulary we invent, the descriptor is the JSON-Schema subset that the provider tool / structured-output endpoints already consume — `{type, properties, required, enum, items, ...}` — exposed through constructors that mirror it. Your set survives almost intact: `arg_enum`, `arg_array`, `arg_string`/`arg_number`/`arg_integer`/`arg_boolean`, with your `arg_record` as `arg_object` (the JSON-Schema name) and `arg_literal` folded into the concrete scalar types. The reason to borrow the vocabulary instead of minting one: it is what OpenAI/Anthropic actually accept, and ellmer round-trips it — `type_*()` serialize to a schema and `ellmer::type_from_schema()` reads one back — so blockr.ai's translation is one call rather than a bespoke per-kind mapper, MCP and a raw tool schema consume the JSON directly, and no harness is privileged. It is also standard and a little more expressive than a hand-rolled record/array set (`anyOf` and friends come for free).

Your actual payoff survives intact, and it is the part I most want to keep: because the descriptor is core-owned and structural rather than opaque, registration can validate it and check the worked example against it. Right now `validate_block_spec()` only confirms an example constructs, so a malformed-but-constructible `conditions = "not-a-list"` passes; against a declared array-of-objects it will not. That is the gap you pointed at, and it closes.

The thing I want to hold off on is the semantic layer — `column`, `expression` as first-class kinds. Two reasons. The boundary is arbitrary: allow `column` and we owe `table`, `model formula`, `aes mapping`, `regex`, `sql fragment`, `file path`, and core becomes the arbiter of which strings are "special" with no natural place to stop. And the value is narrower than it first looks — the real win is "`column` → an enum of the actual upstream columns", which only applies to a bare column *pick*; an expression over columns stays a free string you cannot enumerate, and putting the upstream schema into the model's context already gets most of the way there.

Where the intent goes meanwhile is `description`, which is already in the subset and which core stores without interpreting — "a column in the upstream data", "an R expression", whatever the author wants to say. If blockr.ai later hits a concrete need for the hardened enum-of-columns binding, we add it then as an additive optional annotation shaped around that real consumer; nothing here forecloses it, and unknown keywords are ignored by every consumer in the meantime.

Net: structural descriptor plus example-vs-schema validation now, semantics in `description`, and we revisit a typed annotation when a consumer can name the exact win. Thanks for the offer to prototype — the constructors plus the conformance check are the work, and I have written it up for pickup just below. Push back if you think the column case is more load-bearing than I am giving it credit for.

nbenn:
## Implementation: replace the opaque `type` with a JSON-Schema-subset descriptor

Resolution of the `type`-field discussion — to be implemented in this PR, on `121-block-args`. Rationale is in the reply just above (https://github.com/BristolMyersSquibb/blockr.core/pull/233#issuecomment-4820741814); the validation gap is in @christophsax's review comment (https://github.com/BristolMyersSquibb/blockr.core/pull/233#issuecomment-4817858839).

**Goal.** Replace the opaque `type` on `new_block_arg()` (currently documented as "an `ellmer::type_*`, stored opaquely and not interpreted") with a core-owned, dependency-free descriptor drawn from the JSON-Schema subset that the LLM tool / structured-output APIs accept, and validate worked examples against it.

**Constructors** — new, exported, dependency-free; each returns a plain nested list mirroring JSON Schema:

- `arg_string(description = NULL, required = TRUE)`
- `arg_number()`, `arg_integer()`, `arg_boolean()` — same signature
- `arg_enum(values, description = NULL, required = TRUE)`
- `arg_array(items, description = NULL, required = TRUE)` — `items` is another `arg_*` descriptor
- `arg_object(..., description = NULL, required = TRUE)` — named `arg_*` fields, closed (`additionalProperties: false`)

Names mirror the subset (string / number / integer / boolean / enum / array / object). That is the whole vocabulary — no `const`/`anyOf` builders until something needs them, and no semantic kinds (`column`/`expression`); the reply explains why. The naming prefix (`arg_*` vs the package's `new_*` convention) is open — I lean `arg_*` for call-site readability inside `new_block_arg(type = ...)`; settle it in review.

**Storage.** Keep the descriptor as the plain JSON-Schema list, not an ellmer/S4 object — that is what keeps core dependency-free and lets each consumer bind it directly (blockr.ai via `ellmer::type_from_schema()` or per-node `ellmer::type_*()`; MCP / a raw tool schema consume the JSON as-is). Do not import any consumer's type system into core.

**Validation at registration** — the real new behavior, in `validate_block_spec()` (`R/block-args.R`):

1. Structural — a supplied `type` uses only the supported node kinds and is well-formed.
2. Conformance — for each worked example, every field with a declared `type` must *conform to* its descriptor, not merely construct. The current loop only checks that `do.call(ctor, ...)` succeeds, so a malformed-but-constructible `conditions = "not-a-list"` passes; against a declared array-of-objects it must be rejected.

The conformance check needs a small R-value-against-descriptor validator (R list/vector against the JSON model over the closed node set). Hand-roll it — do not pull in `jsonvalidate` or any other dependency; the node set is small and closed, and dependency-freedom is the entire point of this change.

**Remove / avoid.**

- Drop the opaque-`ellmer::type_*` contract from `new_block_arg()`'s `type` (both the roxygen and the behavior). The `block_arg_type()` accessor stays; it now returns the descriptor list.
- Do not add `column`/`expression` or any other semantic-kind layer. Semantic intent rides in `description`.

**Backward compatibility.** Bare named-character-vector arguments and `character()` normalize exactly as today (no `type`, no new validation). Only specs that opt into `type` are affected.

**Tests.**

- Structural validation rejects an unknown / ill-formed descriptor.
- Conformance rejects a malformed-but-constructible example against a declared `type` — use the `conditions = "not-a-list"` shape, and prove the regression: the test must fail on the pre-change code and pass after. A well-formed example conforms.
- Bare-vector registrations are unaffected.
- (Optional) a representative descriptor round-trips through `ellmer::type_from_schema()` without error — guard with `skip_if_not_installed("ellmer")`; ellmer stays out of `Imports`.

**Scope.** The per-arg `type` slot only. The block-level fields (`details`/`link`/`guidance`/`examples`/`keywords`) and the other accessors are untouched.

PR #245, opened 2026-06-30
Add the reactives class

### LINKED ISSUE #244: A reactive collection with honest membership and computed slots
`reactiveValues` is our dynamic keyed container, but for the variadic block-input bag (`...args`) we keep working around three of its limits. Honest key removal isn't supported, so `trim_rv()` reaches into `.subset2(rv, "impl")` internals to drop a key. Reordering isn't supported, so `reorder_rv()` does the same. And every slot is a *stored* value that something has to push in — a slot can't compute its own value on read — which is why the lazy-pull work fills `...args` with a `needed`-gated writer observer instead of letting each entry pull lazily.

Those map to the empty cell in the reactive taxonomy: `reactiveValues` is to `reactiveVal` (a dynamic collection of *stored* cells) as this proposal is to `reactive()` (a dynamic collection of *computed* cells). Shiny doesn't ship the latter, but it composes cleanly from parts Shiny already exposes.

## Proposal

A small reactive container backed by an environment of active bindings — so slot access runs a closure and is therefore reactive and may compute on read — plus an ordered `keys` `reactiveVal` that is the single source of membership. Membership operations become honest: `makeActiveBinding()` to grow, `rm()` to shrink, a plain reassignment of `keys` to reorder. The read interface mirrors `reactiveValues` (`$`, `[[`, `names()`, `length()`, `as.list()`); writes are explicit helpers rather than `$<-`, because a computed slot is registered with an *expression*, not a value.

```r
new_arg_collection <- function() {
  structure(
    list(store = new.env(parent = emptyenv()),   # active bindings live here
         keys  = reactiveVal(character())),       # ordered; the ONE source of membership
    class = "arg_collection"
  )
}

## write side -- grow / shrink / reorder, all honest
set_arg <- function(x, key, value_fn) {           # value_fn: a 0-arg pull (or reactiveVal read)
  store <- .subset2(x, "store")
  if (exists(key, store, inherits = FALSE)) rm(list = key, envir = store)
  makeActiveBinding(key, value_fn, store)
  k <- .subset2(x, "keys"); if (!key %in% isolate(k())) k(c(isolate(k()), key))
}
drop_arg     <- function(x, key)   { rm(list = key, envir = .subset2(x, "store"))
                                     k <- .subset2(x, "keys"); k(setdiff(isolate(k()), key)) }
reorder_args <- function(x, order) .subset2(x, "keys")(order)

## read side -- mirrors reactiveValues
`$.arg_collection`     <- function(x, name) {
  s <- .subset2(x, "store"); if (exists(name, s, inherits = FALSE)) get(name, envir = s) else NULL
}
`[[.arg_collection`    <- function(x, i) x[[ if (is.numeric(i)) names(x)[i] else i ]]
names.arg_collection   <- function(x) .subset2(x, "keys")()
length.arg_collection  <- function(x) length(.subset2(x, "keys")())
as.list.arg_collection <- function(x, ...) set_names(lapply(names(x), `[[`, x = x), names(x))

## inspection must NOT fire the bindings
print.arg_collection   <- function(x, ...) {
  cat("<arg_collection>", paste(isolate(.subset2(x, "keys")()), collapse = ", "), "\n")
  invisible(x)
}
```

## What it buys

- **Grow / shrink** — `makeActiveBinding` + `rm()` on the backing environment is honest add and remove, retiring `trim_rv()`.
- **Reorder** — order is just the `keys` vector, retiring `reorder_rv()`.
- **Named / unnamed** — `keys` is an ordered character vector, so positional slots are integer ids exactly as `...args` already uses, and `dot_args_names()` keeps working off `names()`.
- **Computed slots** — a slot can be an active binding over a `reactive()` (or a gated pull), so `c$foo` is a value computed on access with a per-slot dependency. That is the variadic value contract (`...args$foo` is a data object, not a reactive) without a writer observer.

## The "don't surprise" constraints

The container has to read like `reactiveValues`, and the design risk lives entirely on that side:

- **Inspection must not fire the bindings.** `print`/`format`/`str` — and console autoprint, `View()` — must read `isolate(names)` only; otherwise inspecting the object runs every slot's computation and silently subscribes the caller. This is the central footgun and the reason such a type isn't a Shiny primitive.
- **`reactiveValuesToList()` won't dispatch on it** (it is `is.reactivevalues`-gated), so the container ships `as.list()` and consumers (e.g. `block_server`'s `dat`) use that.
- **A read of a missing slot returns `NULL` but does not subscribe** to a later add of that key, whereas `reactiveValues` does via per-key dependencies on absent keys. Consumers iterate `names()`, so this is minor, but it should be documented.

## Scope

The computed-slot payoff is moot for our current blocks: `dat` assembles the whole `...args` bag to evaluate the block, so it touches every slot regardless. Per-slot laziness and per-arg invalidation only start to matter once `dat` is itself made lazy (force only the entries the expression references) — a separate change. So the concrete near-term value here is the container ergonomics — honest grow/shrink/reorder and named/unnamed, i.e. deleting `trim_rv()`/`reorder_rv()` and the per-block writer observer — not pull performance.

This does not block the lazy-pull PR, which ships the current shape (`...args` as a `reactiveValues` of values filled by a `needed`-gated writer) and is fine. This is the cleaner long-term container, split out so that PR stays focused.

nbenn:
## Summary

- Adds `reactives`, a dynamic, reference-semantic collection of reactives for variadic block inputs — an environment of active bindings, so a slot computes its value on read. A slot is any 0-arg callable: a `reactive()` (which memoizes), or a `reactiveVal` / plain function (which is also settable).
- Reference type: the store env and keys are shared, so a mutation in one server context (the board) is visible to every other holder (the blocks). `set_reactive` / `drop_reactive` / `reorder_reactives` give honest, in-place grow / shrink / reorder; reorder permutes the keys by position, or by name when every slot is named (else a labelled error), and leaves the store stable.
- Public surface is `list(...)` in insertion order: `as.list(reactives(x, a = y, z))` is `list(x, a = y, z)` and `names()` is `c("", "a", "")`. Unnamed slots get hidden synthetic keys (a real name like `"1"` survives) that nothing consumer-facing shows; position is their address — `[[`, `[` and reorder all take it.
- Write operators delegate to the bindings: `$<-` / `[[<-` push a value through a slot (a `reactiveVal` / function cell sets, a `reactive` cell errors of its own accord), `[<-` re-registers an expression, `[` subsets, `names<-` re-keys.
- Class only — `...args` adopts it in the pull-based-eval work (#211).

Fixes #244


nbenn:
## `...args` positional semantics

Recording the model this PR now implements for variadic `...args`, settled in review.

`...args` behaves as `list(...)`: an ordered run of arguments, each named or positional, where a positional argument's identity is its **absolute position** in the call. A named arg is keyed by its name, a positional arg by its position as a string (`"1"`, `"2"`, …); keys are unique.

**Splice to `list`** — positional args sit at their numbered slots; named args fill the free slots, gaps first then appended past the max, in their own declaration order; emit in ascending position order, unfilled gaps collapsed.

**Construction, two forms (never mixed)** — *inferred*: positional args passed unnamed, each taking its 1-based index in the call with named slots counted (`reactives(x, a = y, z)` → keys `1, a, 3`); *explicit*: positional args given integer-string keys directly, to open a gap (`reactives(\`1\` = x, \`3\` = z, a = y)`).

**Errors at construction** — two args at the same position, or two args with the same name; together these forbid mixing inferred and explicit positional keys.

Behavioral contract (and the test matrix):

```
reactives(x, a = y, z)              →  list(x, a = y, z)
reactives(`1`=x, a=y, `3`=z)        →  list(x, a = y, z)
reactives(`1`=x, `3`=z, a=y)        →  list(x, a = y, z)
reactives(`1`=x, `2`=z, a=y)        →  list(x, z, a = y)
reactives(a = y, x, z)              →  list(a = y, x, z)
reactives(`1`=x, `5`=w, a=y, b=z)   →  list(x, a=y, b=z, w)
reactives(`2`=x, y)                 →  error: position collision
reactives(a=x, a=y)                 →  error: name collision
```

Implementation notes: the ordering — position-sort with named args filling gaps — replaces the old positional-first grouping, and `complete_variadic_inputs` assigns absolute indices (named links counted) rather than `seq_along` over the unnamed. Board ids are stable, so removing a middle link leaves a gap rather than renumbering. The `reactives` container itself is unchanged — it stays a policy-free, order-preserving map.

nbenn:
Scope update after review: this PR is now **just the class** — the positional/named model lives in it (`as.list(reactives(...))` *is* the `list(...)`), but it's adopted nowhere. Re-based onto `main`.

The board-side adoption from the spec above — `complete_variadic_inputs` → absolute index, retiring the `reactiveValues` `...args` bag, the per-block writer observer and `reorder_rv` — moves to #211, where the switch to pull-based eval happens.

PR #251, opened 2026-07-02
Hold blocks with unconnected inputs pending instead of erroring

### LINKED ISSUE #122: Better error when no input exists
If a block is not connected, it shows all kind of messages. A unified, beginner-friendly error message would help.

Image

### LINKED ISSUE #219: Unconnected block evaluates and errors at board startup (is.data.frame(data) is not TRUE)
## What

On a board that contains a downstream block whose inputs are not yet wired — for example a `head_block` with no incoming link — the block evaluates at board startup and logs:

```
[ERROR][blockr.core] is.data.frame(data) is not TRUE
```

The block runs its expression against missing/empty input and raises, which is logged at ERROR on every such startup.

## Question to settle

Should a block with unfulfilled inputs evaluate and raise at ERROR level during startup, or should it stay in a quiescent "pending input" state until its inputs are connected?

This may well be intended — blocks surface their conditions, and an unconnected transform legitimately has no data — in which case this is a "confirm working-as-intended" rather than a fix. But an ERROR-level log on a normal startup of a board that simply has an unwired block is worth a deliberate decision.

## Reproducing

Serve a dock board with an unconnected transform block:

```r
blockr.dock::serve(
  blockr.dock::new_dock_board(
    blocks = c(
      a = blockr.core::new_dataset_block("iris"),
      b = blockr.core::new_head_block()
    )
  ),
  "my_board"
)
```

`b` (head) has no link to `a`, so it evaluates with no data at startup and logs the error. Surfaced via blockr.dock's `edit-link` e2e fixture.

nbenn:
## Summary

- A downstream block with unwired inputs ran its data validator against missing data at board startup, raising and logging an ERROR (`is.data.frame(data) is not TRUE`) on a normal, transient state. Input *presence* is now checked ahead of the validator: a block whose required inputs aren't all connected — or a variadic block below its declared `...args` minimum — is held pending, skipping both validation and evaluation and surfacing a waiting-for-input warning instead of an error.
- The gate is on the evaluation axis only; rendering stays gated on visibility, so a *visible* but unconnected block still renders its UI. A `NULL` input reads as "not present", so a block downstream of a pending one stays pending too — the cascade needs no `req()`.
- `allow_empty_state` gains a list form declaring a variadic block's minimum connected inputs via a `...args = N` entry (e.g. `list(...args = 1)`). `rbind_block` uses it in place of a hand-rolled `stopifnot(length(...args) >= 1L)`; an undeclared variadic block such as `glue_block` keeps its previous zero-input behaviour.
- #219 (the ERROR-level log) and #122 (the pile of raw messages a beginner sees for an unconnected block) are the same root, so both are resolved here.

Fixes #219
Fixes #122

### DIFF
--- a/R/block-server.R
+++ b/R/block-server.R
@@ -17,11 +17,34 @@
 #' i.e. block user inputs and expression), and instantiation of the
 #' `edit_block` module (if passed from the parent scope).
 #'
-#' A block is considered ready for evaluation whenever input data is available
-#' that satisfies validation ([validate_data_inputs()]) and nonempty state
-#' values are available (unless otherwise instructed via `allow_empty_state`
-#' in [new_block()]). Conditions raised during validation and evaluation are
-#' caught and returned in order to be surfaced to the app user.
+#' Each block carries an *eval status* -- one of `dormant`, `waiting`, `unset`,
+#' `failed` or `ready` -- which, together with the orthogonal `visible` flag,
+#' determines its behaviour. The status separates the two input kinds (data
+#' inputs from links, user inputs from `state`) and a genuine failure:
+#' * `dormant` -- not *needed* (neither on screen nor feeding, transitively over
+#'   [board_links()], an on-screen block); inputs stay unfulfilled
+#'   ([shiny::req()] out) and nothing evaluates.
+#' * `waiting` -- needed, but a required *data* input is missing: unconnected,
+#'   below the required number of variadic `...args` inputs (one by default),
+#'   or fed by an upstream block that is not itself `ready` (see
+#'   `allow_empty_state`).
+#' * `unset` -- data inputs are ready, but a required *user* input (`state`
+#'   value) has not been provided (unless permitted by `allow_empty_state`).
+#' * `failed` -- all inputs are present, but the block cannot produce a result:
+#'   the data validator ([validate_data_inputs()]) or the block expression
+#'   raised. The offending condition is surfaced through the block conditions.
+#' * `ready` -- evaluation succeeded and a result (possibly a legitimate `NULL`)
+#'   is available for downstream blocks to consume.
+#'
+#' A block reaches `ready` only once its upstreams have, so an unconnected or
+#' pending block holds its whole downstream chain `waiting` without any of them
+#' evaluating against missing data. Output rendering follows the status: the
+#' block output is shown only while `ready` and cleared otherwise, so a block
+#' leaving `ready` never displays a stale result. While not `ready` the block
+#' surfaces a condition explaining why -- a `status`-phase note for `waiting`
+#' and `unset`, or the raised error for `failed`. Conditions raised during
+#' validation and evaluation are caught and returned to be surfaced to the app
+#' user.
 #'
 #' Block-level user inputs (provided by the expression module) are separated
 #' from output, the behavior of which can be customized via the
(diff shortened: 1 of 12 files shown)


R/block-server.R
```diff
@@ -320,6 +341,54 @@ validate_block_reactive <- function(id, x, dat, cond, sess) {
...
+  }
+
+  if (is.na(block_arity(x))) {
+
+    min_args <- block_min_args(x)
+
+    if (sum(lgl_ply(inp[["...args"]], not_null)) < min_args) {
+      return(
+        paste0(
+          "Waiting for at least ", min_args, " connected input",
```
nbenn: If we have a mix of fixed and variable args, this wording here might be confusing.

R/block-server.R
```diff
@@ -320,6 +341,54 @@ validate_block_reactive <- function(id, x, dat, cond, sess) {
   )
 }
 
+unmet_input_requirements <- function(x, inp) {
+
+  fixed <- block_inputs(x)
+  missing <- fixed[!lgl_ply(inp[fixed], not_null)]
```
nbenn: Can we reliably count on `NULL` always meaning `missing`? I don't think so. You need to come up with a better solution. And if doing so is not straightforward, let's just simplify to something generic like "Input conditions not met" or so.

nbenn (commented):
The main issue here is that I don't think we can agree on a single value to always mean "missing". The current sentinel value `NULL` might be in some instances a valid value.

R/block-server.R
nbenn: Agreed, and the `NULL` sentinel is gone entirely. Each block now derives a four-level eval status (`dormant` / `blocked` / `evaluating` / `ready`), and a downstream block gates on whether its upstreams have reached `ready` — a structural fact — rather than on any input value. So a connected upstream that legitimately produces `NULL` is never mistaken for a missing input. `unmet_input_requirements` and its `not_null` check are removed; validation and evaluation now gate on a board-computed `inputs_ready` (every required slot linked to a `ready` upstream, variadic slots meeting the declared minimum). Readiness propagates, so a pending upstream holds its whole downstream chain `blocked` without any of them validating against absent data.

R/block-server.R
nbenn: Also resolved: a blocked block now surfaces a single generic `Input requirements are not met.` notice instead of naming specific inputs, so there's no fixed-vs-variadic phrasing left to get wrong.

nbenn:
Reworked per review — the `NULL`-as-missing sentinel is gone.

Each block now carries a four-level **eval status** (`dormant` / `blocked` / `evaluating` / `ready`) derived from whether it is needed, whether its inputs are ready (structurally: every required slot linked to a `ready` upstream, variadic slots meeting the declared `...args` minimum), and its own validation/state. Every gating decision is a threshold on that status — pull inputs at `≥ blocked`, evaluate at `≥ evaluating`, feed downstream at `ready` — while rendering stays on the orthogonal visibility bit, so a visible-but-`blocked` block still shows its panel. Because readiness propagates, an unconnected or pending block holds its whole downstream chain `blocked` rather than each erroring against missing data — the cascade you asked to keep, now carried by a status rather than a value.

One edge worth flagging: a `ready` upstream whose *expression* then errors yields a `NULL` result, which a downstream validator rejects as a genuine data condition. That's the one place a `NULL` still reaches a validator, and it's a real failure surfaced as such — distinct from the missing-input case, which never runs a validator at all.

Force-pushed as a single commit; the two inline threads sit on the now-removed `unmet_input_requirements`, replied there.

PR #254, opened 2026-07-02
Order block-server construction by visibility

### LINKED ISSUE #243: Defer block-server instantiation to the needed set (setup_board wires all blocks up front)
## Problem

`setup_board` instantiates every block's server up front: it loops `setup_block` over all `board_blocks` (`R/board-server.R:362`, ~`:378`), inside `observeEvent(TRUE, setup_board(…), once = TRUE)` (~`:160`). So on the first flush all N block servers are wired synchronously, before any visibility information exists. On a large board this blocks first paint by the full per-block setup cost (×99 on blockr.cdex), independent of how many blocks are on screen.

Demand-driven *evaluation* (#204, PR #211 `204-lazy-pull`) defers a block's eval, but the block's server is still instantiated in this loop — so lazy-pull leaves the *instantiation* cost on the startup critical path. (Measured on cedx: even with only the visible closure evaluating, the synchronous all-blocks setup is a large part of the ~148s startup.)

## Direction

Make `setup_board` demand-driven, mirroring the eval gate: instantiate a block's server only when it enters the *needed* set, and instantiate the rest on demand as that set grows (e.g. when a view is opened). The signal already exists — blockr.dock#201 publishes the active view's on-screen block ids to the `visible` channel, and core already derives `needed = upstream_blocks(visible)` for the eval gate — so this consumes the same input.

## Scope and relationship

- This is the **core** half of demand-driven multi-view startup. The **dock** half — deferring off-screen *UI* construction — is a separate dock issue. The two are sibling consumers of the view-active event, not a stack.
- Not covered by dock#215 (post-connect UI assembly); that is dock UI shape, this is server instantiation.
- Extends #204's demand-driven philosophy from eval to instantiation.

## Caveats to work through

- **Mid-session module creation.** Block servers are `moduleServer`s created at session start today; lazy instantiation creates them later. Confirm inputs/outputs bind correctly when a block's UI appears, and that re-opening a view does not re-instantiate (track which servers exist).
- **Upstream closure.** A block that becomes needed must have its upstream instantiated too — the needed set is already the upstream closure, so the set is right; the instantiation order needs to honour it.
- **Plugins / stacks.** `setup_stacks` and the edit/ctrl plugin wiring also run in `setup_board`; decide what defers with the block and what stays eager.

nbenn:
## Summary

- `setup_board()` built every block's server on the first flush, so a large multi-view board paid the full per-block construction cost at first paint regardless of how many blocks were on screen — the instantiation cost that demand-driven evaluation (#204/#211) left on the startup critical path. Construction now follows the same *needed* set (on-screen blocks plus their upstream closure over `board_links()`) as evaluation: the needed set is built synchronously for first paint, and the remaining block servers are built progressively in the background by a self-rescheduling observer, so first paint waits only for what is on screen.
- This is *ordering*, not deferral — every block is still constructed, just off the first-paint path. An unbuilt block is simply *absent* from `rv$blocks` rather than a permanent placeholder, so the enumerating consumers (conditions, notifications, code export) need no changes. Opening a view builds its blocks immediately through the needed-set observer rather than waiting for the background queue; blocks added through the UI, and blocks targeted by an external-control `mod`, build immediately too.
- Serialization is the one consumer that copes with a block still building: it falls back to the block's constructor state (`initial_block_state()`) for any block absent from `rv$blocks`, which is lossless since an unbuilt block has never been touched. The shinytest2 result export is pinned to board order so construction order does not perturb snapshots.
- The background cadence is the `background_construction_delay` option (milliseconds between successive blocks, default 50); setting it to 0 builds every block up front and opts out of the staggering.
- Code export is deliberately *not* touched here. Its output is only correct once every block is ready — a missing link or unset input already yields stale code today — which is a standing correctness concern tracked in #256; gating the export on the per-block ready status from #251 covers the brief startup window as a side effect.

The startup win is realized once a front-end drives the `visible` channel (blockr.dock's producer); standalone core has no producer, so every block is needed and all servers build up front, unchanged.

Fixes #243

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,19 @@
 # blockr.core 0.1.3
 
+* Block-server construction is now ordered by visibility (#243). On a large
+  multi-view board `setup_board()` built every block's server up front, so
+  first paint waited for the full per-block construction cost regardless of
+  how many blocks were on screen. Construction is now prioritized by the
+  same *needed* set (on-screen blocks plus their upstream closure) as
+  evaluation: those blocks are built first and the rest are built
+  progressively in the background, so first paint waits only for what is
+  shown. A block still building is absent from the read-only board handed to
+  plugins, and serialization falls back to its constructor state until it
+  appears. Requires a front-end driving the `visible` channel; without one
+  every block is needed and all servers are built up front, as before. The
+  background cadence is the `background_construction_delay` option
+  (milliseconds between successive blocks, default 50); setting it to 0
+  builds every block up front and opts out of the staggering.
 * Board options contributed by blocks or the block registry -- such as the
   table preview `page_size`, `n_rows` and `filter_rows` -- are now serialized
   and restored along with the board. Previously only options owned by the board
--- a/R/block-server.R
+++ b/R/block-server.R
@@ -65,7 +65,15 @@
 #' fully quiescent: its result reactive, and any observer its expression server
 #' registers on the incoming data, all short-circuit and do nothing. A needed
 #' but off-screen block (one feeding a visible block) evaluates but does not
-#' render. With nothing driving
+#' render. Block-server *construction* is prioritized the same way: the needed
+#' set is instantiated first so that first paint waits only for the on-screen
+#' blocks and their upstreams, and the remaining block servers are built
+#' progressively in the background. Until a block is built it is absent from
+#' the `board$blocks` handed to plugins and callbacks, which simply see it
+#' appear once constructed. The background cadence is set by the
+#' `background_construction_delay` [blockr_option()] (milliseconds between
+#' successive blocks, default 50); a value of 0 disables the staggering and
+#' builds every block up front. With nothing driving
 #' `visible` every block is needed and behaviour is unchanged; the
 #' `gate_visibility` [blockr_option()] (default `TRUE`) turns gating off
 #' entirely.
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -410,11 +410,14 @@ setup_board <- function(rv, blk_ed, blk_ct, stk_mod, args, sess) {
   rv$sources <- list()
   rv$stacks <- list()
 
-  blks <- board_blocks(rv$board)
+  observe(
+    {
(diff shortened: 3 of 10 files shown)


nbenn:
Profiled this together with blockr.dock#289 on blockr.cdex (99 blocks / 12 views). The deferral concept clearly works, but the **default `background_construction_delay = 50` is a net regression here** — the background loop starts before the visible set has painted, so it starves the first paint.

Same stack, only the option changed; time-to-paint of the Setup view:

| `background_construction_delay` | time-to-paint | total settle | R thread during |
|---|---|---|---|
| 50 (default) | **93s** | 126s | pegged ~100% |
| 2000 | **28s** | long low-CPU tail | burst → ~15–25% |

For reference, the pre-deferral stacks were ~148s (old lazy) and ~77s paint / ~90s settle (dock#274 + core main). So at the default this is *slower to interactive* than not deferring at all; at 2000ms it's a big win (~28s, the intended "active view first, rest fills in quietly").

Root cause is timing, not tick size. `construct_remaining_blocks` self-starts the loop on its first observe pass (`started <- TRUE; invalidateLater(...)`), which runs in the same flush where the needed pipeline (`data → prod_adapter → cdisc`) is still evaluating. The needed set finishes evaluating ~20s in, but paint doesn't land until 93s — the whole gap is the background construction of the 95 off-screen servers monopolising the single R thread and starving the dockView layout flush. A large delay only helps because it stops hogging the thread; the real fix is *when* the loop starts.

Desired behaviour: hold the loop until the visible set is constructed **and painted**, then begin ticking at the normal cadence (50ms is fine once it starts at the right moment).

The open question is the start signal — "all set up for the visible blocks." Two candidate sources, your call which fits: a server-side reactive that reads each needed block's result (so it stays pending until the visible set resolves), deferred one flush via `session$onFlushed()` so the paint is sent first; or a client-side "active view painted" signal routed back from the dock. Flagging rather than patching.

(blockr.dock#289 tested clean on its own — only the active view's cards are built, and startup board updates drop to 0, so it eliminates the init churn even without dock#274. No change needed there.)

PR #259, opened 2026-07-02
Gate background construction on the visible set being rendered

### LINKED ISSUE #257: Gate background block construction on the visible set being rendered
Follow-up to #254, to be implemented stacked on it.

## Problem

#254 defers off-screen block-server construction, but `construct_remaining_blocks` **self-starts** the tick loop on its first observe pass (`started <- TRUE; invalidateLater(...)`), in the same flush the visible set is still evaluating and painting. On a large board the background construction of the off-screen servers then saturates the single R thread and starves the first paint.

Measured on blockr.cdex (99 blocks / 12 views), with blockr.dock#289 (build only the active view's cards):

| `background_construction_delay` | time-to-paint | settle |
|---|---|---|
| 50 (default) | 93s | 126s |
| 2000 | 28s | long low-CPU tail |

So at the default it is *slower to interactive* than not deferring (dock#274 + core main was ~77s paint / ~90s settle; old lazy ~148s); a large delay only helps because it stops hogging the thread. The real issue is *when* the loop starts, not the tick size. See #254 comment for the full trace.

## Fix

Hold the background loop until the **visible set is rendered**, then begin ticking at the normal cadence.

The signal is core-owned — no new dock→core message. Two facts make it so: `visible` is already core's own `reactiveVal` (`board_visible` → `rv$visible`), and the block UI/render lifecycle runs through core generics (`insert_block_ui` is `UseMethod`; the render-gate observer is core's). So core can track render state itself.

## Design

- Promote `rv$visible` from binary to a per-block status: `hidden | visible | rendered`. The dock's `visible` channel is unchanged (it still reports the on-screen set → `visible`); core layers `rendered` on top.
- Set `rendered` from core's **render machinery** — the per-block render-gate observer that fires when a visible block's output first paints. Sourcing it there (rather than from `insert_block_ui`) covers both paths uniformly: the static startup active view (rendered via `board_ui` + the dock's `restore_layout`) *and* dynamic inserts. `insert_block_ui` only needs to move a freshly-inserted block into `visible` so it becomes eligible.
- In `construct_remaining_blocks`, replace the self-start with: hold until every currently-`visible` block has reached `rendered`, then start the background cadence. Tick size (default 50ms) is fine once it starts at the right moment.

## Caveat to handle

A visible block parked on missing input may never produce data. Define `rendered` as "the output has flushed at least once" so an empty/prompt state still counts, and/or treat an expr-server that has quiesced (nothing pending) as satisfying the gate — otherwise an input-waiting visible block would hang the loop forever. Quiescence, not data-completion, is the correct trigger.

## Acceptance

On a large multi-view board, the active view paints at roughly its own data-load time (independent of total block count), and off-screen construction starts only afterward and runs at low CPU. No regression versus eager (`background_construction_delay = 0`). Verify on cdex against the numbers above.

nbenn:
## Summary

Stacked on #254 (`243-defer-instantiation`); follow-up per #257.

#254 orders construction visible-first, but the background pass self-started on its first observe pass -- the same flush the visible set is still evaluating and painting -- so on a large board off-screen construction saturates the single R thread and starves first paint (cdex: 93s to paint at the default delay vs. ~77s without deferring). The lever is *when* the loop starts, not the tick size.

- Adds a per-block `rendered` signal, sourced from core's render machinery (`block_rendered_observer`) and returned by `block_server()`. It flips once the result has settled while the block is on screen -- quiescence, not paint -- so an input-waiting visible block showing a prompt still counts and can't hang the gate. It reads `res()` only when on screen, exactly where `output_render_observer` already forces it, so off-screen laziness is unchanged.
- `construct_remaining_blocks()` now holds until `visible_set_rendered(rv)` -- every on-screen block constructed and `rendered` -- then hands off to the background ticker. The status `hidden | visible | rendered` is the front-end's `visible` set (unchanged) with core's per-block `rendered` layered on top.
- The gate and the ticker are separate observers so the ticker is driven only by `invalidateLater`; folding them let a later `visible` change retrigger construction, which the existing "opening a view" test caught.

`background_construction_delay = 0` still builds every block up front.

The acceptance -- active view paints at ~its own data-load time regardless of total block count -- is a wall-clock property, verified on cdex against #257's numbers; unit coverage is the gate predicate and the quiescence trigger, both with regression teeth.

Fixes #257

Revdep pin (temporary — drop once blockr.dock#306 merges): core#259 reshapes the `visible` channel into a per-block status map, which the released dock does not yet write, so the blockr.dock revdep must build against the coupled adaptation PR rather than dock `main`.

```deps
BristolMyersSquibb/blockr.dock#306
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,17 @@
   background cadence is the `background_construction_delay` option
   (milliseconds between successive blocks, default 50); setting it to 0
   builds every block up front and opts out of the staggering.
+* The background block-construction pass (#243) now waits for the front-end to
+  arrange the on-screen blocks before it starts (#257). It previously
+  self-started in the same flush that first paint was still evaluating, so on a
+  large multi-view board the off-screen construction saturated the single R
+  thread and starved the active view's arrangement. The `visible` write-channel
+  now carries a per-block status (off screen / on screen / rendered) instead of
+  a flat id set; the on-screen set that feeds the eval/needed gate is derived
+  from it unchanged, and the background pass holds until every on-screen block
+  is reported rendered. With nothing driving `visible` the board is all on
+  screen and ready, as before, and `background_construction_delay = 0` still
+  builds every block up front.
 * Board options contributed by blocks or the block registry -- such as the
   table preview `page_size`, `n_rows` and `filter_rows` -- are now serialized
   and restored along with the board. Previously only options owned by the board
--- a/R/block-server.R
+++ b/R/block-server.R
@@ -52,8 +52,9 @@
 #' control rendering of outputs.
 #'
 #' When a front-end (such as blockr.dock) drives the `visible` write-channel
-#' that [board_server()] hands to the board callback, naming the block IDs
-#' currently on screen, evaluation and rendering are gated on visibility.
+#' that [board_server()] hands to the board callback, reporting each block's
+#' visibility status (off screen, on screen, or rendered into its view),
+#' evaluation and rendering are gated on visibility.
 #' Rendering is gated on plain visibility: the render observer is suspended
 #' while a block is off screen and resumed once it is on screen, starting
 #' suspended so nothing renders before the front-end first reports. Evaluation
@@ -68,7 +69,10 @@
 #' render. Block-server *construction* is prioritized the same way: the needed
 #' set is instantiated first so that first paint waits only for the on-screen
 #' blocks and their upstreams, and the remaining block servers are built
-#' progressively in the background. Until a block is built it is absent from
+#' progressively in the background. That background pass holds until the
+#' front-end reports every on-screen block as rendered (arranged into its
+#' view), so it never competes with first paint. Until a block is built it is
+#' absent from
 #' the `board$blocks` handed to plugins and callbacks, which simply see it
 #' appear once constructed. The background cadence is set by the
 #' `background_construction_delay` [blockr_option()] (milliseconds between
@@ -508,7 +512,8 @@ render_gate_observer <- function(id, board, render_obs, sess) {
 
   observe(
     {
-      do_render <- isTRUE(board$visible) || id %in% board$visible
+      do_render <- isTRUE(board$visible) ||
+        isTRUE(board$visible[id] %in% c("required", "rendered"))
 
       if (do_render) render_obs$resume() else render_obs$suspend()
 
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -95,7 +95,7 @@ board_server.board <- function(id, x, plugins = board_plugins(x),
           cur <- if (isTRUE(rv$visible)) {
             TRUE
           } else {
-            upstream_blocks(rv$visible, rv$board)
+            upstream_blocks(on_screen_blocks(rv$visible), rv$board)
           }
 
           old <- isolate(rv$needed())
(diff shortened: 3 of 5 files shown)


nbenn:
Reshape needed, coordinated with blockr.dock#289 — the gate here keys on the wrong signal.

**The problem (measured on cdex):** `block_rendered_observer` flips `rendered` when `res()` settles — the block's *result* (~20s) — but the panel isn't arranged into the visible view until the dock's `input$initialized` observer runs (blockr.dock `board-server.R:594`: `restore_layout()` + the `show_block_panel()` loop). The background construction, released at result-quiescence, then starves that observer to ~85s. So: **95s time-to-paint / 131s settle at the default cadence**, vs ~28s when the background is held off (verified with `background_construction_delay = 2000`). The gate must wait for the dock's *arrangement*, not the result settling.

**The reshape:** make `visible` a **multi-level per-block status** instead of a flat id set, and let the dock drive the top level. Levels: `hidden` (off-screen / absent) | `required` (on-screen, arrangement pending) | `rendered` (arranged). Provisional names.

Piggyback on `visible` — not a sibling `rendered` channel. One source of truth; the `rendered ⊆ required` invariant (and its reset when a block re-enters `visible` on a view switch) is structural rather than a cross-set constraint core has to police; and forward-only means there's no compat cost to reshaping the existing channel. (Every desync loop we've hit this session is two pieces of state chasing each other — one channel shrinks that surface.)

**Core side:**
- `board_visible` (`board-server.R:123`) carries the status map (id → level) instead of an id vector; still passed as `visible` in the `cb_args` list (`:149`).
- Derive **on-screen** = ids at `required` ∪ `rendered` → feeds the existing eval/`needed` gate unchanged (`rv$visible`, `upstream_blocks(...)`).
- Derive **ready** = every on-screen block at `rendered` → `visible_set_rendered` / the `construct_remaining_blocks` start gate holds until ready.
- **Delete `block_rendered_observer`** (block-server.R:460) — `rendered` now comes from the dock, not from `res()`.
- **No-producer default**: absent/unset producer ⇒ ready = TRUE and on-screen = all, mirroring today's `NULL → all-visible` (a non-dock board, or startup before the first dock report, must not hold the gate forever).

The `required`→`rendered` writes are the dock side, in BristolMyersSquibb/blockr.dock#289.

nbenn:
Reshaped as described — `visible` now carries a per-block status map and the gate keys on the dock's arrangement rather than `res()`. Pushed to `257-render-gate`.

Core side:

- `board_visible` carries `id → level` (`required` / `rendered`; off screen is absent, or `"hidden"`). The observer at `board-server.R:123` derives **on-screen** = `on_screen_blocks(status)` (the `required ∪ rendered` ids) into `rv$visible`, feeding the eval/`needed` gate and the render gate unchanged, and **ready** = `visible_set_rendered(status)` (every on-screen id at `rendered`) into `rv$visible_rendered`.
- `construct_remaining_blocks` holds until `isTRUE(rv$visible_rendered)`, then hands off to the background ticker (the gate/ticker split stays, so a later view switch can't re-trigger it).
- `block_rendered_observer` and the `rendered` reactive on `block_server()`'s return are deleted — `rendered` comes from the front-end now, not `res()`.
- No producer / gating off / before the first report ⇒ `rv$visible = TRUE`, `rv$visible_rendered = TRUE`, mirroring the old `NULL → all-visible`.

On the startup window: no race, because the dock already seeds `visible` synchronously in the callback (`blockr.dock board-server.R:67`, with the "else core's all-visible default evaluates every block at startup" comment) before the first flush. So in the dock case `rv$visible_rendered` starts `FALSE` and holds; #289 seeds the initial `required` the same way and promotes to `rendered` after `restore_layout` / `show_block_panel`. The `NULL → TRUE` default only applies to genuine non-dock boards, where `needed = all` leaves the background nothing to build anyway.

Side benefit of routing the signal through the channel: the hold is now directly testable. `test-visibility-gating.R` seeds `b` at `required`, asserts the background stays put across a 5s elapse, then promotes it to `rendered` and asserts it drains — and it fails against an ungated start, so the hold has teeth (my earlier `res()`-quiescence version couldn't exercise the hold in a synchronous test at all).

Contract for #289: `visible(c(id = "required" | "rendered", ...))`, blocks off screen absent. Level names kept as you proposed (provisional — trivial to rename). Full suite green locally; CI stays idle until the stack retargets to `main`.

PR #267, opened 2026-07-07
Preserve link/stack/block string fields matching JSON Inf/NA tokens

### LINKED ISSUE #136: Link input "Inf" becomes numeric after JSON round-trip
It's a bit niche, but took us a while to debug on BMS prod: Links with input `"Inf"` fail to restore because jsonlite converts the string `"Inf"` to numeric `Inf` during deserialization.

Repro: `new_link("a", "b", "Inf")` → save board → restore → error

Why would you want to name them `Inf`? I don't know...

nbenn:
## Summary

- `jsonlite` decodes the JSON tokens `"Inf"`, `"-Inf"`, `"NaN"` and `"NA"` back to the numeric/logical values they usually encode, so a contractually-character field equal to one of them came back non-character through a `write_json`/`read_json` cycle and aborted the board restore (found in prod with a link input `"Inf"`).
- Coerce the affected fields back to character on deserialization via a shared `restore_chr()` helper: link `from`/`to`/`input`, stack `name` and member block IDs, and block `block_name`.
- Block IDs need no fix — they serialize as JSON object keys, and keys are never type-coerced on read.
- Out of scope by design: block state and board-option *values*, which are genuinely type-ambiguous (a value may legitimately be numeric `Inf` or the string `"Inf"`). No global `toJSON`/`fromJSON` setting disambiguates them — `serializeJSON` would, at the cost of verbose, opaque JSON and a breaking format change — so that remains a documented limitation of the interchange format.

Fixes #136

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,6 +5,11 @@
   and restored along with the board. Previously only options owned by the board
   itself (e.g. its name) survived a save/restore cycle, so customizations to
   these block-sourced options were silently lost (#146).
+* A board with a `link` input, a `stack` name or member block ID, or a block
+  name that is exactly `"Inf"`, `"-Inf"`, `"NaN"` or `"NA"` now survives a
+  save/restore cycle. `jsonlite` parses these JSON tokens back to the
+  numeric/logical values they encode, so the field previously returned as a
+  non-character value and aborted the restore (#136).
 * Blocks now carry an eval status -- `dormant`, `waiting`, `unset`, `failed` or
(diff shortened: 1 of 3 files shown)


nbenn:
Closing unmerged — decided to wontfix #136 (jsonlite interchange limitation; the case barely occurs and any real fix has a real cost). This field-by-field coercion papers over the jsonlite limitation unprincipledly and targets a case that essentially doesn't arise, so it's not worth shipping. Full rationale on #136; the seam-level annotation approach is the path if we ever revisit.

Issue #269, opened 2026-07-10
Support `background_construction_delay = Inf` to defer construction until needed


nbenn:
## Idea

The `background_construction_delay` knob added in #254 (for #243) controls ordered startup:

- `0` — build every block up front (opt out of ordering).
- finite positive `N` — build the needed set first, then the rest in the background, one block every `N` ms.

The natural other end of the spectrum is `Inf`: never build a block proactively — construct it only when it enters the *needed* set (on screen, or feeding something on screen). A block that is never visited is never built, so a large multi-view board only ever pays for the blocks the user actually opens.

This is the pure demand-driven deferral behaviour (the first cut of #243, before it settled on ordering). Exposing it as `Inf` rather than the default keeps ordering as the out-of-the-box behaviour while letting a deployment opt into "only build what's viewed" — worthwhile when most views on a big board are never opened in a session.

## Implementation sketch

Small. `construct_remaining_blocks()` already branches on the delay (`<= 0` → build everything up front, else start the staggered background observer). Add an `is.infinite()` branch that skips the background scheduler entirely — the needed-set observer already builds blocks on demand as views open, so `Inf` just means "don't run the background fill". `as.numeric()` already accepts `Inf`.

## Interaction to resolve

With `Inf`, off-screen blocks that are never visited stay unbuilt and are therefore absent from `rv$eval`. The code-export gate `code_export_ready()` (#256, merged as #258) holds on `setequal(names(rv$eval), board_block_ids(...))`, so "Show code" would report "board not ready" until every block has been visited at least once. Either that is acceptable (the user visits the views before exporting) or `Inf` mode pairs with a "materialize all blocks on export" path. Serialization needs no extra work — an unbuilt block already serializes from its constructor state.

PR #274, opened 2026-07-13
Gate blocks on separate required and visible channels

### LINKED ISSUE #273: Board callback should provide `visible` as a reactive, not write a reactiveVal
`board_server()` gates block-server construction, evaluation and output rendering on a per-block visibility status (`required` / `rendered`, off-screen omitted) that a front-end drives on the `visible` channel. Today that channel is a **push**: core creates `board_visible <- reactiveVal()`, hands it to the board callback as the `visible` argument, and the front-end runs an observer that materialises its layout into that reactiveVal (with its own dedup) for core to read back.

The front-end's layout is the single source of "what is on screen." The push makes core hold a materialised *copy* of it that the front-end has to keep in sync with an observer, and — because it is a written channel — it invites the write-back / ping-pong loops we have fought hard to eliminate on the layout in blockr.dock.

Flip it to a **pull**: the board callback *returns* a `visible` reactive whose body derives the status from the front-end's own layout, and core reads it. Core keeps its `NULL -> all-visible` default and the `gate_visibility` option, reading the reactive instead of a written value:

```r
# core, in the visibility observer:
vis <- callback_visible()   # a reactive the callback returned
rv$visible <- if (is.null(vis) || !gate_visibility()) TRUE else vis
```

Now the layout is the one source (core reads a live derivation — nothing to materialise, dedup, or keep in sync), and it is one-directional by construction: core reads, so it structurally cannot write back to the front-end's layout state. That is strictly stronger than "be careful not to write back."

The interface change is the point, and it is breaking for the callback contract: `visible` goes from an *input* (a reactiveVal core creates and the front-end writes) to an *output* (a reactive the callback returns and core reads). blockr.dock then drops its report-visible observer and its dedup and just returns the derivation.

One caveat: `rendered` (the client has painted a view) is a stateful event, so the front-end's `visible` reactive reads a small per-view latch for it — that stays entirely on the front-end side; core still just consumes the resulting status map.

Surfaced while building the card-deferral / `visible`-channel work in BristolMyersSquibb/blockr.dock#289.

nbenn:
## Summary

Block-visibility gating moves from one core-owned `visible` push channel to **two per-block channels that core and the front-end share**: `required` (the *cause* — which blocks are needed) and `visible` (the *fulfillment* — which view each block is painted into). Core creates both, owns their per-block slot lifecycle, validates them, and reads them to drive building and evaluation; the front-end (blockr.dock) writes them.

This fixes the root problem behind #273: `needed` was single-sourced from `visible`, so a second writer — e.g. code export marking a block needed — was clobbered within a flush. Splitting cause from fulfillment gives `needed` a genuine multi-writer source and keeps `visible` as the front-end's paint report.

## The two channels

Core bundles them as `vis = list(required = <env>, visible = <env>)` and threads it as an **explicit argument** through block construction — never on the read-only board. Each channel is an **environment of per-block `reactiveVal`s** (`id → reactiveVal`) rather than a `reactiveValues`, because a `reactiveValues` can't shrink (`rv[[id]] <- NULL` leaves the key) whereas `rm(list = id, envir = …)` removes a slot cleanly, and an environment is reference-shared so both value writes and slot add/remove reach every holder. Writers set a slot by calling its reactiveVal — `visibility$required[[id]](TRUE)`, `visibility$visible[[id]](view)`.

- **`required[[id]]`** — `TRUE` (needed now: build + evaluate), `FALSE` (was required, kept built but dormant), `NA` (never required: unbuilt). The one shared object takes writes from any source — dock, plugins, core-side features like code export.
- **`visible[[id]]`** — the view id the block is painted into, or `NA_character_`. Core reads only whether it's a non-`NA` string; the value itself is opaque to core.

Core seeds a block's slots (`add_vis_slots`) at construction and on block-add — before any of that block's plugins run — drops them on remove (`rm_vis_slots`), and a max-priority observer hard-errors on any off-contract write, so readers can trust the types.

The `NULL` → all-visible default and the `gate_visibility` option are unchanged: a board with no `required` set (any non-dock board) builds and renders everything, as before.

## Breaking change — callback contract

The board callback now receives `visibility` (the `vis` bundle) instead of a `visible` reactiveVal to write, and reports by calling the per-block reactiveVals rather than pushing a status map. blockr.dock adopts this in the linked PR.

```deps
BristolMyersSquibb/blockr.dock#289
BristolMyersSquibb/blockr.dag@core274-pin
BristolMyersSquibb/blockr.assistant@core274-pin
```

## Notes

- **Version / NEWS:** first PR into freshly-released `main` — patch-bumped to 0.1.4 with a breaking-change entry.
- **Merge order:** the queue-only revdep builds the downstream stack against this PR, so it must resolve to the adopted dock. dock#289 adopts the new channels and pins `core@273-visible-pull`; blockr.dag and blockr.assistant transitively depend on dock, so throwaway `core274-pin` branches repin them to dock#289 for the revdep. The deps block points the revdep at all three. Once this lands, those throwaway pins are dropped and dock#289 repins to `main`.

Fixes #273

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: blockr.core
 Title: Graphical Web-Framework for Data Manipulation and Visualization
-Version: 0.1.3
+Version: 0.1.4
 Authors@R: 
     c(person(given = "Nicolas",
              family = "Bennett",
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,10 @@
+# blockr.core 0.1.4
+
+* The board callback now gates block construction, evaluation and rendering
+  through two per-block `reactiveValues` it receives as `visibility` --
+  `required` (which blocks are needed) and `visible` (which are on screen) --
+  in place of the single `visible` write-channel. Breaking for front-ends.
+
 # blockr.core 0.1.3
 
 * Block-server construction is now ordered by visibility -- on-screen blocks
--- a/R/block-server.R
+++ b/R/block-server.R
@@ -18,8 +18,9 @@
 #' `edit_block` module (if passed from the parent scope).
 #'
 #' Each block carries an *eval status* -- one of `dormant`, `waiting`, `unset`,
-#' `failed` or `ready` -- which, together with the orthogonal `visible` flag,
-#' determines its behaviour. The status separates the two input kinds (data
+#' `failed` or `ready` -- which, together with its orthogonal front-end
+#' visibility, determines its behaviour. The status separates the two input
+#' kinds (data
 #' inputs from links, user inputs from `state`) and a genuine failure:
 #' * `dormant` -- not *needed* (neither on screen nor feeding, transitively over
 #'   [board_links()], an on-screen block); inputs stay unfulfilled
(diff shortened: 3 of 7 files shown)


nbenn:
## Reframe request: externalize `needed` (the cause), not `visible` (the effect)

TL;DR — coming out of the #269 code-export work, I think #274 is externalizing the wrong channel. The thing that wants to be a pulled, multi-writer channel is `needed` (the cause); `visible` is the effect and should stay the front-end's arrangement report. This also dissolves the `rv$materialize` flag #270 currently carries.

### Why

#269 (code export under `background_construction_delay = Inf`) needs "Show code" to mark the whole board *needed*, so a genuinely-unconfigured off-screen block holds the export back instead of silently emitting broken code. Concretely, an off-screen `merge` with no `by` is `unset`; build-only leaves it `dormant`, and `code_export_ready` then passes and emits `merge(.(x), .(y), by = character(0))`. Making it *needed* surfaces the `unset` and gates the export — that's the correct behaviour, and it's a general code-export hole, not `Inf`-specific.

But `needed` is single-sourced: one observer derives it from `visible` and nothing else can contribute. A back-end requirement written onto `visible` (or straight onto `rv$needed`) is clobbered by the next front-end report within one flush. Measured: require-all → `merge` `unset`, `code_export_ready` `FALSE` (correct); one stray `rendered` report lands → `needed` reverts, `merge` back to `dormant`, `code_export_ready` `TRUE`, the broken code is exportable again. So the fix isn't a flag — it's that `needed` needs to be a channel multiple parties can write, and #274 is the natural place to do that.

### The model

`needed` = **cause** channel, externalized and multi-writer. Per-block `TRUE` (needed now) / `FALSE` (ever needed, not now) / *absent* (never) — presence + boolean, deliberately not `NA`. The front-end declares its active-view blocks; the back-end (code export, and whatever comes next) declares its set; core reads the **union**.

`visible` = **effect** channel, shape unchanged: the arrangement outcome the front-end reports (`required` → `rendered`, off-screen absent), with "a `required` block always resolves to `rendered`/parked" as its contract. It's the *result* of satisfying a display need, not the need itself.

`eval` (`dormant`/…/`ready`) is untouched — still derived from needed-now.

### The tri-state is the build/eval split, for free

- construction target = **present** (`TRUE` or `FALSE`) — ever-needed, monotonic → `needed_block_ids(rv)` becomes `names(rv$needed)`.
- eval / `dormant` gate = **`TRUE`** — needed-now → `block_needed(rv, id)` becomes `isTRUE(rv$needed[[id]])`.

A block the back-end needed and then released is `FALSE`: still built, back to `dormant`. That's exactly #269's requirement with no `materialize` state — that flag was me faking a second writer onto a single-writer channel.

### What this changes in #274

Keep the push→pull direction, but flip `needed`, not `visible`:

- the callback returns/contributes `needed` (a reactive core reads for the front-end's declared set); the back-end contributes `needed` core-side; core unions them.
- `visible` stays the front-end-reported effect (`required`/`rendered`); the render gate and the #257 background-construction gate keep reading it unchanged.

Consumers barely move: build → `names(needed)`, eval → `isTRUE(needed[[id]])`, gates → `visible`.

Still breaking for blockr.dock — it now *declares* `needed` (its view's blocks) rather than reporting `visible` status — so the same merge-order / revdep-pin dance the current #274 already describes.

### Sequencing

Land the reframed #274 first. Then #269/#270 collapses to a few lines: the back-end writes all-`TRUE` while the code modal is open and withdraws on close, replacing the `rv$materialize` flag. Happy to take the dock side once the core shape here is settled.

nbenn:
Merge is gated on the `merge_group` revdep, which builds blockr.dock against this PR: the dock callback has to adopt the new `required` / `visible` channels first, and dock `main` (and dock#289 at `02725f1`) still write the old `visible` write-channel.

blockr.dock#289 is being reworked to adopt them. Once it's ready, this pin moves into the PR **body** — where the revdep's parse-deps step reads it — and #274 is enqueued so the queue builds the adopted dock:

```deps
BristolMyersSquibb/blockr.dock#289
```

nbenn:
Adopting this in blockr.dock (#289) surfaced one ordering issue in `apply_core_board_update`.

The dock single-stores its build ledger — which block cards are in the offcanvas DOM pool — in the `required` channel: a card is built exactly when its `required` slot is non-NA (`ever_required`). So `insert_block_ui` marks a newly added block's card built by writing `required[[id]](FALSE)`. For that write to land, the slot has to exist when the plugin runs.

In the `upd$blocks$add` branch, `add_vis_slots()` currently runs *after* `insert_block_ui`, and it installs a fresh `reactiveVal(NA)` per id — so the plugin's write either targets a slot that doesn't exist yet or is clobbered a line later. Hoisting the seed above the plugins fixes it:

```r
if (length(upd$blocks$add)) {

  log_debug("adding block{?s} {names(upd$blocks$add)}")

  add_vis_slots(vis, names(upd$blocks$add))   # was below construct_blocks

  do.call(insert_block_ui, ...)
  board_blocks(rv$board) <- c(board_blocks(rv$board), upd$blocks$add)
  construct_blocks(names(upd$blocks$add), rv, ...)
}
```

This makes the add path honor the invariant construction already holds — every block's slots are seeded before any block plugin runs (the `add_vis_slots(vis, isolate(board_block_ids(rv$board)))` at construction). The add branch is the one place a block plugin can currently run against an unseeded slot. `add_vis_slots` only creates reactiveVals keyed by id, so hoisting it above the `board_blocks` write is safe.

Without it, a dock block-add leaves the new card outside the built ledger (its `required` stays NA after the clobber), so the next view switch rebuilds the card — a doubled-DOM regression. Flagging here rather than patching, since the PR is yours.

nbenn:
Good catch — fixed in `86638b3` (folded into the reframe commit). `add_vis_slots(vis, names(upd$blocks$add))` now runs at the very top of the `upd$blocks$add` branch, above `insert_block_ui` and the `board_blocks(rv$board) <- …` write, so the add path holds the same invariant construction already does: every new block's `required` / `visible` slots exist before any of its block plugins run. A dock plugin that marks a card built via `required[[id]](FALSE)` inside `insert_block_ui` now lands on a live slot instead of one `add_vis_slots` later clobbers with a fresh `reactiveVal(NA)`.

Core-side the reorder is behavior-preserving — none of core's own plugins write `required` in `insert_block_ui`, so nothing here distinguished the two orders; the full suite is unchanged at 1739/0. The doubled-DOM regression only becomes observable once a plugin writes `required` during the add, so the guarding e2e belongs with #289's adoption rather than here.

Force-pushed to `273-visible-pull`.

PR #283, opened 2026-07-16
Drop un-deserializable blocks instead of aborting board load

### LINKED ISSUE #264: Could we make deser more robust?
Errors in deserialization remain one of the bigger annoyances. 

While they may occur for any reason, could the framework not just drop offending blocks, extensions, etc, rather than crash?

nbenn:
## Summary

- A single block that can't be deserialized -- its constructor or package is no longer available, its payload won't reconstruct, or the round-trip class check fails (`block_deser_class_error`) -- currently aborts the *entire* board restore. That is the recurring annoyance this issue describes, and specifically covers a board referencing a block *type* a package has since retired or consolidated.
- `blockr_deser()` for `blocks` gains an `on_error` argument, defaulting to `blockr_option("deser_on_error", "abort")`. `"abort"` is the current behaviour, unchanged; a deployment opts into `"drop"` via `options(blockr.deser_on_error = "drop")` or the `BLOCKR_DESER_ON_ERROR` environment variable.
- In `"drop"` mode each block deserializes behind a `tryCatch`; a block that fails any of the three modes is omitted (with a `log_warn` naming the block and the reason) and the rest of the board loads.
- Dropping a block cascades: `blockr_deser.board()` prunes the links and stacks that reference an *actually-dropped* block before `new_board()` validates, so the load doesn't simply re-abort with a dangling-reference error. `validate_board()` is left untouched -- a reference to a block that was never present (genuine corruption, not a drop) still aborts.
- Scope is blocks plus the link/stack cascade in core; extensions can adopt the same `on_error` protocol later.

Fixes #264

### DIFF
--- a/R/utils-serdes.R
+++ b/R/utils-serdes.R
@@ -254,9 +254,21 @@ blockr_ser.stacks <- function(x, ...) {
   )
 }
 
+#' @param on_error How to handle a block that cannot be deserialized -- its
+#' constructor (or the package providing it) is unavailable, its payload cannot
+#' be reconstructed, or the round-trip class check fails. `"abort"` (the
+#' default) propagates the error, while `"drop"` omits the offending block
+#' (emitting a warning) and continues; board deserialization then prunes any
+#' links and stacks that reference a dropped block, so a board referencing a
+#' since-retired or no-longer-installed block type still loads. Board
+#' deserialization resolves `on_error` once and threads it to the contained
+#' blocks. The default is sourced from `blockr_option("deser_on_error",
+#' "abort")`, letting a deployment opt into dropping via
+#' `options(blockr.deser_on_error = "drop")` or the `BLOCKR_DESER_ON_ERROR`
+#' environment variable.
 #' @rdname blockr_ser
 #' @export
-blockr_deser <- function(x, ...) {
+blockr_deser <- function(x, ..., on_error = NULL) {
   UseMethod("blockr_deser")
 }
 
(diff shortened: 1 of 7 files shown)


R/utils-serdes.R
```diff
@@ -307,26 +307,77 @@ blockr_deser.block <- function(x, data, ...) {
...
+#' blocks. The default is sourced from `blockr_option("deser_on_error",
+#' "abort")`, letting a deployment opt into dropping via
+#' `options(blockr.deser_on_error = "drop")` or the `BLOCKR_DESER_ON_ERROR`
+#' environment variable.
 #' @rdname blockr_ser
 #' @export
-blockr_deser.blocks <- function(x, data, ...) {
-  as_blocks(
-    lapply(data[["payload"]], blockr_deser)
+blockr_deser.blocks <- function(x, data, on_error = NULL, ...) {
```
nbenn: Should we not make this arg part of the generic itself?

R/utils-serdes.R
```diff
@@ -307,26 +307,77 @@ blockr_deser.block <- function(x, data, ...) {
...
 
   stopifnot(
     all(c("constructor", "payload") %in% names(data))
   )
 
+  on_error <- resolve_deser_on_error(on_error)
+
   ctor <- blockr_deser(data[["constructor"]])
 
+  parts <- lapply(data[["payload"]], blockr_deser, on_error = on_error, ...)
```
nbenn: How about passing `on_error` after the dots?

R/utils-serdes.R
```diff
@@ -340,6 +391,37 @@ blockr_deser.board <- function(x, data, ...) {
   res
 }
 
+drop_block_refs <- function(parts, blocks) {
```
nbenn, line 394: Do we not have utils for this? We are doing similar things when preprocessing `update()`s.

nbenn (commented):
Seen comments inline.

R/utils-serdes.R
nbenn: Done in 069492a. `on_error` is now an argument of the `blockr_deser()` generic itself (after `...`), so it's documented once as the deserialization error policy. `blockr_deser.list()` passes it straight through; `board` and `blocks` declare it after their own `...` to act on it, and the other methods absorb it via `...`.

R/utils-serdes.R
nbenn: Done in 069492a — `on_error` now sits after `...` in the generic and the method signatures, and after the dots at the call site too (`lapply(data[["payload"]], blockr_deser, ..., on_error = on_error)`).

R/utils-serdes.R
nbenn: There wasn't a shared helper yet — the predicate `links[links$from %in% ids | links$to %in% ids]` was inlined at four sites (`rm_blocks`, `destroy_rm_blocks`, `augment_board_update`, and this cascade). Extracted it as `links_incident(links, ids)` and reused it at all four (069492a). `augment_board_update()` itself isn't reusable wholesale since it's bound to the update-delta shape, so this shares just the atom.

PR #286, opened 2026-07-16
Force live/nm in freeze_state_view() so state views stop aliasing

### LINKED ISSUE #285: Block state views alias onto the last state field, so saved boards store wrong state
### Summary

`freeze_state_view()` binds every one of a block's state fields to the **last**
field. Any block with two or more state fields reports the last field's value for
all of them, and `serialize_board()` writes that out — so a user's edit to any
field but the last is silently dropped from the saved board, and the last field's
value leaks into its siblings.

This is on `main` (11fff26). Regression from **70acf18 "Freeze block inputs via a
per-block frozen channel"** (2026-07-14).

### Root cause

`R/block-server.R`:

```r
for (nm in views) {
  state[[nm]] <- freeze_state_view(live[[nm]], nm, frozen, snapshot, sess)
}

freeze_state_view <- function(live, nm, frozen, snapshot, sess) {
  reactive(
    if (frozen()) snapshot()[[nm]] else reval_if(live),
    domain = sess
  )
}
```

`reactive()` defers its body, and `live` / `nm` arrive as unforced promises. They
are therefore only resolved on the **first read**, in the caller's frame — where
the `for` loop has long since finished and `nm` is the last element of `views`.
Every view ends up reading `live[[last]]`.

It only bites where `visibility` is non-null, i.e. any served board — the freeze
path is what builds these views.

### Minimal reproducible example

Runs against `main` with no app, no browser:

```r
library(blockr.core)
library(shiny)

# a block with TWO state fields
new_probe <- function() new_transform_block(
  function(id, data) {
    moduleServer(id, function(input, output, session) {
      first_rv  <- reactiveVal(FALSE)
      second_rv <- reactiveVal(FALSE)
      observeEvent(input$first,  first_rv(input$first))
      observeEvent(input$second, second_rv(input$second))
      list(
        expr  = reactive(quote(identity(data))),
        state = list(first = first_rv, second = second_rv)
      )
    })
  },
  function(id) tagList(),
  class = "probe",
  block_metadata = FALSE
)

probe <- new_probe()

vis <- list(
  required = new.env(parent = emptyenv()),
  visible  = new.env(parent = emptyenv()),
  frozen   = new.env(parent = emptyenv())
)
blockr.core:::add_vis_slots(vis, "p")

testServer(
  blockr.core:::get_s3_method("block_server", probe),
  {
    session$flushReact()
    scope <- session$makeScope("expr")

    scope$setInputs(first = TRUE)          # set ONLY the first field
    session$flushReact()
    cat("set first=TRUE  -> first =", blockr.core:::reval_if(session$returned$state$first),
        "| second =", blockr.core:::reval_if(session$returned$state$second), "\n")

    scope$setInputs(first = FALSE, second = TRUE)   # set ONLY the second field
    session$flushReact()
    cat("set second=TRUE -> first =", blockr.core:::reval_if(session$returned$state$first),
        "| second =", blockr.core:::reval_if(session$returned$state$second), "\n")
  },
  args = list(
    x = probe,
    data = list(data = reactiveVal(datasets::BOD)),
    block_id = "p",
    visibility = vis
  )
)
```

On `main`:

```
set first=TRUE  -> first = FALSE | second = FALSE     # the edit is LOST
set second=TRUE -> first = TRUE  | second = TRUE      # second's value leaks into first
```

Expected (and what the fix gives):

```
set first=TRUE  -> first = TRUE  | second = FALSE
set second=TRUE -> first = FALSE | second = TRUE
```

### How it surfaced

`blockr.dm`'s `cdisc_dm_block` "forgetting" its **Set CDISC keys** checkbox: tick it,
save the board, reload, unticked. `set_keys` aliases to `dedup_cols` (the last state
field), so ticking `set_keys` saved nothing — while ticking only `dedup_cols` saved
**both** as `TRUE`.

Confirmed end-to-end in a live app (chromote, real trusted click, dumping what
`serialize_board()` received): Shiny had `set_keys = true` and the block's own
reactiveVal was `TRUE`, but serialization captured `FALSE`. The block is innocent —
the aliased view is what gets serialized.

The fix below has since also been confirmed to resolve the issue in production.

### Fix

`force(live)` / `force(nm)` at the top of `freeze_state_view()`, plus a regression
test asserting each exposed view tracks its own value (fails on both assertions
without the fix).

Branch: `fix/freeze-state-view-promise-aliasing`

### Ruled out by experiment

- **ser/deser round-trip is fine.** Real saved JSON restores `set_keys: TRUE`; a
  full serialize → JSON → deserialize cycle preserves both fields.
- **`observeEvent(..., ignoreInit = TRUE)` in the block is NOT the cause.** It looks
  guilty (shiny's `bindEvent()` really does test `ignoreInit` before `ignoreNULL`, so
  `testServer` shows the first tick swallowed) but removing it changed nothing live —
  `testServer` has no startup flush, so that repro is an artifact.
- The block's reactiveVal updates correctly (traced: `rv now = TRUE`).

christophsax:
Fixes #285. Root cause, MRE and verification in the issue.

`reactive()` defers its body, so the `live` / `nm` promises passed from the `for (nm in views)` loop were only resolved on first read — after the loop had finished — binding every state view to the last field. `force()` on both resolves them in the right frame.

- `R/block-server.R`: `force(live)` / `force(nm)` at the top of `freeze_state_view()`.
- `tests/testthat/test-input-freeze.R`: regression test asserting each exposed view tracks its own value. Fails on both assertions without the fix.

Verified: test-input-freeze 26 passed / 0 failed with the fix (2 failures on unfixed main). Also green: block-server (73), board-server (131), plugin-serdes (45), utils-serdes (23). The fix has additionally been confirmed to resolve the issue in production.

### DIFF
--- a/R/block-server.R
+++ b/R/block-server.R
@@ -678,6 +678,15 @@ freeze_exposed_state <- function(state, x, frozen, sess) {
 
 freeze_state_view <- function(live, nm, frozen, snapshot, sess) {
 
+  # `live` and `nm` MUST be forced here. The caller passes them from a
+  # `for (nm in views)` loop, and reactive() defers its body: without these
+  # forces the promises are only resolved on the first read, in the caller's
+  # frame, where the loop has long since finished -- so every view would bind
+  # to the LAST element of `views`. That aliased all of a block's state fields
+  # onto one value, which is what serialize_board() then wrote out.
+  force(live)
+  force(nm)
+
   reactive(
     if (frozen()) snapshot()[[nm]] else reval_if(live),
     domain = sess
--- a/R/block-server.R
+++ b/R/block-server.R
@@ -651,7 +651,7 @@ freeze_exposed_state <- function(state, x, frozen, sess) {
   }
 
   ctrl <- intersect(fn_names, external_ctrl_vars(x))
-  views <- setdiff(fn_names, ctrl)
+  readonly <- setdiff(fn_names, ctrl)
 
   live <- state[fn_names]
 
@@ -669,21 +669,21 @@ freeze_exposed_state <- function(state, x, frozen, sess) {
     hold_ctrl_state(live[ctrl], frozen, snapshot, sess)
   }
 
-  for (nm in views) {
-    state[[nm]] <- freeze_state_view(live[[nm]], nm, frozen, snapshot, sess)
+  for (nm in readonly) {
+    state[[nm]] <- freeze_readonly_field(live[[nm]], nm, frozen, snapshot, sess)
   }
 
   state
 }
 
-freeze_state_view <- function(live, nm, frozen, snapshot, sess) {
+freeze_readonly_field <- function(live, nm, frozen, snapshot, sess) {
 
   # `live` and `nm` MUST be forced here. The caller passes them from a
-  # `for (nm in views)` loop, and reactive() defers its body: without these
+  # `for (nm in readonly)` loop, and reactive() defers its body: without these
   # forces the promises are only resolved on the first read, in the caller's
-  # frame, where the loop has long since finished -- so every view would bind
-  # to the LAST element of `views`. That aliased all of a block's state fields
-  # onto one value, which is what serialize_board() then wrote out.
+  # frame, where the loop has long since finished -- so every field would bind
+  # to the LAST element of `readonly`. That aliased all of a block's state
+  # fields onto one value, which is what serialize_board() then wrote out.
   force(live)
   force(nm)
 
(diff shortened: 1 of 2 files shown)


nbenn:
Pushed a small follow-up (2492d08) renaming `freeze_state_view` → `freeze_readonly_field` and the local `views` → `readonly`.

The name `view` collided with the layout view already in this codebase — the tab/panel a block renders into (`visibility$visible[[id]](view)`, "active view", "view switch" in `board-server.R`) — so inside `block-server.R` the word meant two unrelated things. The renamed pair instead names the property that actually distinguishes these fields from `ctrl`: they have no external control channel, so they're read-only.

Pure rename, no behaviour change — full `test_local` (1807 passed) and lint green. Squash it into your fix commit if you'd rather keep the PR single-commit.

PR #292, opened 2026-07-21
Route block result previews through a tabular display seam

### LINKED ISSUE #129: Table display
Rework table display. I have a draft but let's discuss this carefully first. 

Image

What do we want to show?
- Types

nbenn:
## Summary

- Data, parser and transform blocks hardcoded a DT preview of their result. The only way to swap it (e.g. for blockr.ui's HTML table) was for a downstream package to re-register core's `block_output` / `block_ui` / `block_render_trigger` / `board_options` methods on core's own classes — foreign-method clobbering that is load-order dependent, admits a single winner, and lets the container, renderer and option set drift out of sync.
- Introduce a **`tabular_display`**: one S3 object bundling all four facets, resolved from the `blockr.tabular_display` option (`blockr_option()`). A preview is now chosen without touching core's S3 method tables, and because the container, renderer, trigger and options all dispatch on one object they cannot desync.
- The default is now **`minimal_display`** — a dependency-free `print()` of the top `n_rows` rows in a `verbatimTextOutput`. **`dt_display`** restores the previous paginated, searchable DT table. Downstream packages add their own (e.g. an HTML table) by defining the four `tabular_*` generics on their own display sub-class and having users opt in via the option.
- Board options follow the active display rather than a hardcoded superset: `minimal` contributes `n_rows`; `dt` contributes `n_rows` / `page_size` / `filter_rows`. DT stays a hard dependency — it still backs the links and stacks editors.

Breaking: every data / parser / transform block across the stack now previews with `minimal_display` unless a display is selected.

| Want | Set |
|---|---|
| DT table (previous default) | `options(blockr.tabular_display = blockr.core::dt_display)` |
| blockr.ui HTML table | `options(blockr.tabular_display = blockr.ui::html_table_display)` — after the blockr.ui follow-up |

Follow-up (blockr.ui): register `html_table_display` on the seam and migrate apps off the now-inert `blockr.html_table_preview` option.

Fixes #129


R/data-block.R
```diff
@@ -21,26 +21,26 @@ new_data_block <- function(server, ui, class, ctor = sys.parent(), ...) {
...
+    tabular_ui(tabular_display(), NS(id, "result"))
   )
 }
 
 #' @export
-#' @include utils-dt.R
-block_render_trigger.data_block <- dt_render_trigger
+#' @include tabular-display.R
+block_render_trigger.data_block <- function(x, session = get_session()) {
+  tabular_render_trigger(tabular_display(), session)
```
nbenn: Given that we have `tabular_output()`, `tabular_ui()`, how about we make this `tabular_trigger()` for symmetry reasons?

R/data-block.R
```diff
@@ -21,26 +21,26 @@ new_data_block <- function(server, ui, class, ctor = sys.parent(), ...) {
...
+}
 
 #' @export
 board_options.data_block <- function(x, ...) {
   combine_board_options(
-    new_n_rows_option(...),
-    new_page_size_option(...),
-    new_filter_rows_option(...),
+    tabular_board_options(tabular_display(), ...),
```
nbenn: And `tabluar_options()`?

_pkgdown.yml
```diff
@@ -79,6 +79,10 @@ reference:
     - new_transform_block
     - new_plot_block
     - new_text_block
+- subtitle: Result display
```
nbenn: Do we really need a separate section for this? And if anything, this would group more with utils, no?

R/tabular-display.R
```diff
@@ -0,0 +1,209 @@
...
+  UseMethod("tabular_board_options")
+}
+
+#' @export
+tabular_ui.minimal_display <- function(x, id) {
+  verbatimTextOutput(id)
+}
+
+#' @export
+tabular_output.minimal_display <- function(x, result, block, session) {
```
nbenn, line 107: How does printing of cols work here wrt available width? *If* we want to offer this "minimal" display option out of the box, tibble printing might be more suitable, no?

nbenn (commented):
See inline comments.

R/data-block.R
nbenn: Done — renamed to `tabular_trigger` (generic, both methods, and the three block delegators).

R/data-block.R
nbenn: Done — `tabular_board_options` → `tabular_options` too.

_pkgdown.yml
nbenn: Dropped the standalone section and moved `tabular-display` under Utilities → Miscellaneous, next to `blockr_option` (which is how the active display is selected). The strict pkgdown index just needs the topic listed *somewhere*, not its own heading — agreed it didn't warrant one.

R/tabular-display.R
nbenn: Agreed, switched to it. Base `print.data.frame` wraps a wide frame into stacked column-blocks (re-printing the row index each time), unreadable for anything ADaM-wide; `renderPrint(tibble::as_tibble(...))` truncates to the console width with a `… N more variables` footer and adds the type row — the preview we actually want, and it collects lazy tables cheaply via `head()`.

One flag: this adds `tibble` to Imports. It isn't a hard dep today (only transitive via dplyr), so a standalone core install wouldn't have it. `tibble` is tiny and ubiquitous so I think it earns its place for a sensible default, but if you'd rather keep core lean the alternative is base `print` with a bumped `width` so columns scroll horizontally instead of wrapping. Leaving this thread open for your call on the dep.

nbenn: Done — moved `tibble` to Suggests. The minimal display formats via tibble when it's installed and falls back to base `print()` otherwise (guarded by an internal `tibble_available()` so both branches are tested). The `head()` stays inside `renderPrint`, so a lazy/remote table still isn't pulled when the preview is off-screen.

PR #301, opened 2026-07-24
Re-resolve a downstream input when its upstream builds later

### LINKED ISSUE #298: Finite `background_construction_delay` latches a downstream input at `NULL` when it runs before the upstream is built
A finite, non-zero `background_construction_delay` — the default, `50` — can latch a downstream block's data input at `NULL` for the whole session, because `upstream_result()` reads the upstream server under `isolate()` and so never re-fires when that server is built a moment later. The three construction regimes in [`construct_remaining_blocks()`](https://github.com/BristolMyersSquibb/blockr.core/blob/8bbbcfe73ad0266615fd72970744a027d7213dea/R/board-server.R#L526) diverge here: `0` (eager) and `Inf` (demand-only) are unaffected; only the staggered one-block-per-flush path opens the window. Since `50` is the default, the stock configuration is the affected one.

## Symptom

Reproduced on the `blockr.cdex` full board (`run_cedx()`, 88 blocks) at the default delay, on a cold process's first session. The shared upstream `cdisc` block ("CDISC dm (keys + dedup)") comes up with a persistent red `Input must be a dm object` banner and the navbar construction spinner never clears. The board is wired `data -> prod_adapter -> cdisc -> global_filter -> ` every view's flatten block, so a dead `cdisc` starves the entire app — every downstream view is empty. The browser console also fills with Shiny output-state desync errors on the block's `status_indicator` output ("progress message ... in an unexpected state of: running"; "is recalculating ... state 'idle'"; "has been recalculated ... 'idle'").

It reproduced on both cold starts I ran at `delay = 50`. The same board at `delay = 0` and at `delay = Inf` came up clean (zero errors) on a fresh cold process each. A warm second session against an already-running `delay = 50` process also came up clean, so the trigger is timing-sensitive — the first construction on a fresh process pays per-block lazy-load costs that widen the window.

## Root cause

I instrumented [`upstream_result()`](https://github.com/BristolMyersSquibb/blockr.core/blob/8bbbcfe73ad0266615fd72970744a027d7213dea/R/board-server.R#L929) and the data validator in [`validate_block_reactive()`](https://github.com/BristolMyersSquibb/blockr.core/blob/8bbbcfe73ad0266615fd72970744a027d7213dea/R/block-server.R#L447). At the failure, `cdisc`'s input reactive logs exactly once:

```
INSTR upstream_result to=cdisc key=data from=prod_adapter srv=<NULL> result=NULL
INSTR inputs for cdisc: data=NULL
[ERROR] Input must be a dm object
```

The upstream `prod_adapter` block server is evaluated about two seconds *later*. So `cdisc`'s input reactive first executes while `prod_adapter`'s server is not yet registered in `rv$blocks`: staggered construction builds one block per flush, and a needed (on-screen) block like `cdisc` can have its input reactive run before its hidden upstream `prod_adapter` has been constructed. The link is resolved correctly (`from = prod_adapter`), but `rv$blocks[["prod_adapter"]][["server"]]` is `NULL` at that instant, so the input resolves to `NULL`.

The defect is [line 942](https://github.com/BristolMyersSquibb/blockr.core/blob/8bbbcfe73ad0266615fd72970744a027d7213dea/R/board-server.R#L942):

```r
srv <- if (is.null(from)) NULL else isolate(rv$blocks[[from]])[["server"]]
```

The `[[from]]` access sits *inside* `isolate()`, so the reactive takes no dependency on `rv$blocks[[from]]`. It caches `NULL` and is never invalidated when `prod_adapter`'s server registers a moment later. The reactive runs exactly once (I counted) and the input stays `NULL` for the rest of the session.

Compare [`input_ready()`](https://github.com/BristolMyersSquibb/blockr.core/blob/8bbbcfe73ad0266615fd72970744a027d7213dea/R/board-server.R#L795-L802), which handles the same construction race correctly and documents why:

```r
not_null(from) && identical(reval_if(isolate(rv$eval)[[from]]), "ready")
```

There the `isolate()` wraps only the *container* (`rv$eval`) and `[[from]]` stays outside it, so the per-key dependency survives and "a downstream still wakes when its upstream's status reactive is installed at construction." `upstream_result()` isolates the whole `rv$blocks[[from]]` access instead, dropping that per-key dependency. The two are then inconsistent: `input_ready()` re-fires and flips `cdisc`'s `inputs_ready` to `TRUE`, but `upstream_result()` does not re-fire, so the data it feeds the now-"ready" block stays `NULL`.

`delay = 0` runs `construct_blocks()` over the whole board in one synchronous, topo-ordered pass, and `delay = Inf` builds each demand batch together — both register every upstream before any downstream input reactive runs, so `srv` is never `NULL`. Only the finite path interleaves construction with flushes and opens the window.

## Suggested direction

Would mirroring the `input_ready()` pattern in `upstream_result()` — isolating only the container and keeping the per-key dependency on `rv$blocks[[from]]` — be the right fix? That way the input reactive re-pulls when the upstream server is installed, instead of latching the `NULL` it saw first. I'm happy to open a PR with a regression test that drives staggered construction of a downstream-before-upstream pair and asserts the input recovers.

## Secondary, separate follow-up

The `dm` blocks in `blockr.dm` hard-`stop("Input must be a dm object")` on a non-`dm` input (`R/cdisc-dm-block.R`, and the same guard in `dm-select` / `dm-write` / `dm-nested-view`). That turns what should be a transient "waiting for upstream" state into a persistent red error that then cascades. Even with the core fix above, a `req()`-style soft validation there would be more robust — a transient `NULL` input during construction would surface as "waiting" rather than a hard failure. That belongs in a separate `blockr.dm` issue; noting it here only as the reason this particular board surfaces the core bug so loudly.

## Reproduction

The `blockr.cdex` full board at the default `background_construction_delay`, cold process, first browser connection — no code change needed, it is the default. I have a standalone harness (isolated library, a parametrized launcher keyed on `CDEX_DELAY` / `CDEX_PORT`, and a headless `chromote` driver) and can share it or fold it into a minimal `testServer` regression.

nbenn:
## Summary

- A finite `background_construction_delay` (the default, `50`) could latch a downstream block's data input at `NULL` for the whole session. Under staggered construction a block's input reactive can run before its upstream is registered in `rv$blocks`, and `upstream_result()` read the upstream server under `isolate()`, so it never re-fired once that server registered a moment later. On the `blockr.cdex` board this starved the shared `cdisc` block and every downstream view.
- `upstream_result()` now takes a per-key dependency on the upstream's `rv$eval` slot — installed at the upstream's construction — so the input re-resolves the server when the upstream appears, instead of caching the first `NULL`. This mirrors `input_ready()`, which already depends on that slot to wake a downstream when its upstream's status installs.
- The wake rides `rv$eval[[from]]` rather than `rv$blocks[[from]]` because `rv$blocks` is a plain list with no per-key reactivity; `rv$blocks` stays isolated so the input does not re-fire on every block's construction (`delay = 0` / `Inf` were already immune, building each set in one topo-ordered pass).
- Regression test drives a downstream input that runs before its upstream is built, then installs the upstream mirroring `construct_block`'s order and asserts the input recovers — it fails on the unfixed code (latched `NULL`) and passes with the fix.

Fixes #298

### DIFF
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -939,7 +939,18 @@ upstream_result <- function(key, src_rv, rv, to) {
 
       from <- src_rv[[key]]
 
-      srv <- if (is.null(from)) NULL else isolate(rv$blocks[[from]])[["server"]]
+      srv <- if (is.null(from)) {
+        NULL
+      } else {
+        # Depend on the upstream's rv$eval slot -- installed when it is
+        # constructed -- so a downstream whose input runs before its upstream
+        # is registered re-resolves the server once that upstream appears,
+        # rather than latching the NULL it first saw. Mirrors input_ready();
+        # rv$blocks stays isolated to avoid a whole-container dependency that
+        # would re-fire every input on every block's construction.
+        rv$eval[[from]]
+        isolate(rv$blocks[[from]])[["server"]]
+      }
 
       if (is.null(srv)) NULL else srv$result()
     }
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -939,18 +939,19 @@ upstream_result <- function(key, src_rv, rv, to) {
 
       from <- src_rv[[key]]
 
-      srv <- if (is.null(from)) {
-        NULL
-      } else {
-        # Depend on the upstream's rv$eval slot -- installed when it is
-        # constructed -- so a downstream whose input runs before its upstream
-        # is registered re-resolves the server once that upstream appears,
-        # rather than latching the NULL it first saw. Mirrors input_ready();
-        # rv$blocks stays isolated to avoid a whole-container dependency that
-        # would re-fire every input on every block's construction.
-        rv$eval[[from]]
-        isolate(rv$blocks[[from]])[["server"]]
+      if (is.null(from)) {
+        return(NULL)
       }
+
+      # Depend on the upstream's rv$eval slot -- installed when it is
+      # constructed -- so a downstream whose input runs before its upstream
+      # is registered re-resolves the server once that upstream appears,
+      # rather than latching the NULL it first saw. Mirrors input_ready();
+      # rv$blocks stays isolated to avoid a whole-container dependency that
+      # would re-fire every input on every block's construction.
+      rv$eval[[from]]
+
+      srv <- isolate(rv$blocks[[from]])[["server"]]
 
       if (is.null(srv)) NULL else srv$result()
     }
--- a/NEWS.md
+++ b/NEWS.md
@@ -69,6 +69,12 @@
   Each tick's pacing delay now begins once the just-built block has flushed
   rather than while its reactive graph is still flushing, so pending user input
   is serviced within one tick instead of behind the entire backlog (#276).
+* With a finite `background_construction_delay` (the default), a downstream
+  block's data input no longer latches at `NULL` when its input reactive runs
+  before the upstream is built. The input read the upstream server under
+  `isolate()` and never re-resolved once that server registered a moment
+  later; it now re-fires when the upstream is constructed, so the block picks
+  up its data instead of starving for the rest of the session (#298).
 * Captured block conditions are no longer glue-interpolated when logged, so a
   block whose warning or error text contains braces -- e.g. the `{summary_fun}`
   / `{data}` placeholders in `tidyr::pivot_wider()`'s duplicate-value warning --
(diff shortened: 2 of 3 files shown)


R/board-server.R
```diff
@@ -939,7 +939,18 @@ upstream_result <- function(key, src_rv, rv, to) {
 
       from <- src_rv[[key]]
 
-      srv <- if (is.null(from)) NULL else isolate(rv$blocks[[from]])[["server"]]
+      srv <- if (is.null(from)) {
+        NULL
```
nbenn: Why don't we just return early here and move the rest our of the else branch?

nbenn (commented):
Lgtm

R/board-server.R
nbenn: Good call — done in a1c5241. Early `return(NULL)` for the no-link case, and the eval-slot dependency plus server read are lifted out of the else branch.

PR #302, opened 2026-07-24
Degrade metadata-less blocks to defaults instead of aborting

### LINKED ISSUE #299: Constructing an unregistered block yields a metadata-less block (warning), which then aborts hard downstream
## What happens

`unregister_blocks()` removes a block's registry entry, but its constructor stays exported and keeps working. The block it returns has **no `block_metadata` attribute** — signalled only by a warning. Any later consumer that reads that metadata aborts hard.

```r
library(blockr.core)

b1 <- new_head_block(n = 15L)
is.list(attr(b1, "block_metadata"))   # TRUE

unregister_blocks("head_block")

b2 <- new_head_block(n = 15L)
#> Warning: No block metadata available for block head_block.
is.list(attr(b2, "block_metadata"))   # FALSE

blockr.core:::block_metadata_record(b2)
#> Error: Block head_block carries no metadata.
```

The abort is in `block_metadata_record.block()` (`R/block-meta.R:153-159`), which `blockr_abort()`s whenever `attr(x, "block_metadata")` is not a list.

## Why it matters

The gap between the two steps is the problem: construction degrades quietly, consumption is fatal, and the two can be far apart. Curating the registry is a normal thing for an app to do — the blockr.cloud gallery demos all unregister core blocks down to `dataset` + `glue` — so it is easy to unregister a block the board itself still constructs and get no usable signal until an unrelated consumer explodes.

In our case the consumer was `blockr.dag`'s g6 init, which calls `block_metadata()` for a node **icon**:

```
init_g6 -> g6_from_board -> g6_data_from_board -> g6_nodes_from_blocks -> blks_icon
  -> block_metadata.default -> block_metadata_record.block -> blockr_abort
Error: Block head_block carries no metadata.
```

That killed `board_server` outright and the app rendered a permanently blank dock — no error surfaced in the UI, only in the container log. A cosmetic icon lookup took down the whole board.

## Suggested direction

Any one of these would have caught it far earlier:

1. **Fail at construction**, not at consumption — if a constructor cannot attach metadata, error there, where the cause is still visible. (Or keep metadata on the block independent of the registry, since the constructor knows its own class.)
2. **Make metadata lookup non-fatal** for cosmetic consumers — return `NULL` / a default record and let callers decide, rather than `blockr_abort()` from inside an icon lookup.
3. Failing both, have `unregister_blocks()` warn when it drops an entry whose constructor is still reachable.

Filing rather than fixing since this straddles core and dag — happy to test whichever direction you pick.

Found on blockr.core 0.1.4 / blockr.dag 0.1.2 (current mains, 2026-07-22).

nbenn:
## Summary

- A block constructed while its class is absent from the registry carries no `block_metadata` attribute. Every metadata consumer funnels through `block_metadata_record.block()`, which `blockr_abort()`ed on the missing attribute — so a cosmetic lookup (the reported case: a `blockr.dag` node icon) took down the whole `board_server`, leaving a blank dock with the error only in the container log.
- Fix the single choke point: `block_metadata_record.block()` now returns an empty record (`list()`) instead of aborting. Every downstream consumer already fills display defaults for missing fields (`icon_column`, `name_column`, `scalar_meta`), so a metadata-less block now tabulates with defaults — identical to a block built with `block_metadata = list()`, which already worked.
- This is the issue's direction 2 (non-fatal consumption). I avoided direction 1 (abort at construction: it would break legitimately constructing an unregistered block, and still crash any board that does) and direction 3 (warn on unregister: a band-aid). Construction still warns, so the early signal is preserved.
- Adds a regression test, verified failing against the unfixed code: a metadata-less block, both alone and mixed into a `blocks()` collection, now tabulates with defaults instead of aborting.

Fixes #299

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -123,6 +123,13 @@
   and forward to the new ones, so existing
   `register_block(arguments = new_block_args(...))` call sites keep working;
   migrate them to the `arg_spec` family (#295).
+* A block constructed for a class with no registry entry is now imputed a
+  class-derived default metadata record at construction (name from the class,
+  default category and icon) instead of being left without one. It still warns,
+  but the block is then self-describing: `block_metadata()` and the
+  `block_meta_*()` accessors report those defaults rather than a metadata read
+  aborting -- so a cosmetic lookup can no longer take down a board whose
+  registry has been curated (#299).
 
 # blockr.core 0.1.3
 
--- a/R/block-class.R
+++ b/R/block-class.R
@@ -220,10 +220,13 @@ new_block <- function(server, ui, class, ctor = sys.parent(), ctor_pkg = NULL,
       block_metadata <- c(list(id = uid), block_metadata)
 
     } else if (!isFALSE(block_metadata)) {
+
       blockr_warn(
-        "No block metadata available for block {class[1L]}.",
+        "No registry entry for block {class[1L]}; using default metadata.",
         class = "missing_block_metadata"
       )
+
+      block_metadata <- block_default_metadata(class)
     }
   }
 
--- a/R/block-meta.R
+++ b/R/block-meta.R
@@ -10,7 +10,9 @@
 #' (`block_meta_name()`, `block_meta_guidance()`, ...) returning that attribute
 #' for a single block. Missing fields are filled with display defaults in the
 #' data.frame; the getters instead return the stored value (or `NA` / an empty
-#' value).
+#' value). A block constructed for a class with no registry entry carries a
+#' class-derived default record, imputed at construction, so accessors degrade
+#' to those defaults rather than signalling a missing record.
 #'
 #' @param x A `block`, a `blocks` collection, a `block_registry_entry` or a
 #'   registry ID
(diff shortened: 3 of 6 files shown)


nbenn:
The merge queue ejected this PR once on a transient dependency-download flake, not a code failure — re-enqueuing.

On the `merge_group` ref, the `revdep / revdep (blockr.dock)` job died during dependency install: `Failed to download htmltools 0.5.9 (x86_64-pc-linux-gnu-ubuntu-24.04)` → `Error in file(con, "rb") : cannot open the connection` for `.../src/contrib/x86_64-pc-linux-gnu-ubuntu-24.04/4.6/htmltools_0.5.9.tar.gz` (`No such file or directory`). blockr.dock's own tests never ran. Failed run: https://github.com/BristolMyersSquibb/blockr.core/actions/runs/30078823455.

Evidence it was a per-job PPM/CDN blip, not a persistent missing binary: the sibling revdep jobs (`blockr.assistant`, `blockr.session`, `blockr.dag`) downloaded htmltools fine in the same run, and htmltools 0.5.9 — the current CRAN version — is now present on PPM (source tarball `HTTP 200`; noble binary index lists `htmltools 0.5.9`). All real checks passed: the four-platform `ci / check` matrix and `ci / check-all`.

Issue #321, opened 2026-08-13
Evaluation demand should be one multi-owner set, not two channels


nbenn:
Core does not care *why* a block needs evaluating. Today the demand is nonetheless split in two: a front-end writes `visibility$required[[id]]`, while everything else goes through the `evaluate` / `require` board update components added in #319. That split is incidental rather than designed — the request components were added under #318 without rewriting a channel that had just stabilized — and it means the same demand is expressed two ways depending on who is asking.

## What the `required` axis is actually doing

Three jobs in one tri-state:

1. **Evaluation demand** — `required[[id]]` is `TRUE`, so the block joins the eval set.
2. **Construction demand** — the slot is non-`NA` while `FALSE`: build this block even though nothing needs it evaluated. The dock writes exactly this in `mark_cards_built()`, so a later view switch is fast.
3. **Gating activation** — `gating_active()` is `has_required()`, "has any slot ever been non-`NA`".

Only the first is demand in the sense the eval set cares about, and it is the one that should become a single multi-owner set with the front-end as one owner among several. The other two are separate concerns that happen to share a slot.

## What must not be folded in

**Gating activation.** It is not a demand at all; it is a front-end declaring that it will drive visibility, and it flips the whole board from evaluate-everything to evaluate-only-what-is-needed. Fold a consumer's claim into the same set and an extension asking about one block on an ungated board turns gating on and parks every other block. That cannot happen today: the #319 request sets do not feed `gating_active()`, so on an ungated board a request is a no-op because everything is needed already. A unified set needs this as an explicit capability declaration rather than something inferred from slot history.

**Paint acknowledgement.** The `visible` axis is the client reporting what it has painted, not a demand. The `required_fulfilled()` predicate compares the two — `all(lgl_ply(required_now(vis$required), block_visible, vis))` — and the background constructor holds on it, so collapsing them degenerates that predicate to trivially true and removes the guard that keeps background construction from competing with first paint. The lead of demand over acknowledgement is also what lets evaluation overlap first paint rather than serialize behind it.

**Construction demand** needs somewhere to live. It is a real and distinct demand, just one that only the front-end has had so far. Note that the "have I built this card" bit is currently duplicated across both axes — `mark_cards_built()` writes `visible FALSE` *and* `required FALSE`, and the ledger then reads `!is.na(visible)` — which is a symptom of construction demand living in the demand slot. Expressing it separately removes the duplication.

## Evidence the split has already cost something

[...]

PR #322, opened 2026-08-14
Build the board for code export instead of evaluating it

### LINKED ISSUE #320: Showing code latches every block required, turning a lazy board eager
Clicking "Show code" once converts a lazily-evaluating board into an eagerly-evaluating one for the rest of the session, and takes the front-end's gating state with it.

The export plugin needs every block built and honestly statused, and gets there through `require_all_blocks()`, which writes `visibility$required[[id]](TRUE)` for every block on the board. Nothing ever writes them back. Worse, that write lands on the slot the front-end owns: a block the front-end had parked with `required[[id]](FALSE)` is overwritten, and since the front-end only rewrites that slot on a view change, its intent is lost rather than merely supplemented.

Measured under `testServer` with gating active, parking `b` and then calling `require_all_blocks()` once:

```
parked      : b required = FALSE | status = dormant
after export: b required = TRUE  | status = ready
5 flushes on: b required = TRUE  | status = ready
```

The needed set stays `a b` indefinitely, so `b` re-evaluates on every upstream change from then on, off screen, with nothing to release it.

## Fix

Move the plugin onto the `require` board update component added in #319, which is a request set core owns rather than the front-end's channel: claim on open with `update(list(require = list(add = board_block_ids(board$board))))`, release with the matching `rm`. A leaked claim then only leaves blocks needed — the front-end's own `required` state is untouched, and releasing restores it exactly.

The open question is where the release goes, since `modalDialog()` fires no server-side close event. Either the plugin gains one (a `hidden.bs.modal` binding writing an input), or export switches to a one-off `evaluate` request plus a verdict captured when the request drains — `code_export_state()` reads live status today, which would read every block as `dormant` again once a one-off completes, so that path needs the guard reworked rather than left as is.

nbenn:
## Summary

- Showing the code claimed the whole board by writing `visibility$required[[id]](TRUE)` for every block. Nothing ever wrote those back, and that slot belongs to the front-end, so a block parked with `required[[id]](FALSE)` was overwritten rather than supplemented: one "Show code" turned a lazily evaluating board into an eagerly evaluating one for the rest of the session, with nothing left to release it.
- The export asks for construction alone, through the `construct` component from #333. It assembles the script from block expressions and needs none of their results — `lang` (returned as `expr`) is not gated on `state_ready()` — so a board that defers its off-screen blocks now stays deferred while the code is shown. Nothing is claimed and nothing is released.
- Whether export is safe is read from what each block reports about itself rather than from eval status: `state_ready()` for user inputs that were never set, and the block's error conditions for its last run. Both are readable without putting anything in the eval set, and conditions outlive the evaluation that raised them. Eval status could not serve here, since a block reports `dormant` whenever nothing needs it, which says nothing about whether it works.
- A block that has never run reports neither, so the modal offers a one-off `evaluate` for the board. That leaves every block dormant again, but has each of them report what it found.

Worth spelling out, since it drove the design: an unconfigured merge block exports `merge(a, b, by = character(0), ...)`, which does not error when run — it cross joins, 6 x 6 into 36 rows. The harm is a silently wrong result rather than a crash, and `state_ready` is what catches it.

Breaking for a front-end supplying its own `generate_code_server()`:

| before | after |
| --- | --- |
| `generate_code_server(id, board, visibility = NULL, ...)` | `generate_code_server(id, board, update, ...)` |

Nothing downstream is affected in practice — blockr.dock reuses core's plugin server as is, replacing only where its UI is placed.

Fixes #320

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,20 @@
   request joins neither the eval set nor the front-end's `required` channel, so
   it cannot turn a lazily evaluating board into an eagerly evaluating one
   (#333).
+* Showing the generated code no longer writes the front-end's `required`
+  channel. A block parked with `required[[id]](FALSE)` was overwritten and
+  never restored, so a single "Show code" turned a lazily evaluating board
+  into an eagerly evaluating one for the rest of the session, with nothing
+  left to release it. The export asks for construction alone through the
+  `construct` board update component, since the script is assembled from
+  block expressions and needs its blocks built rather than run: a board that
+  defers its off-screen blocks now stays deferred while the code is shown.
+  Export is held back by what a block reports about itself -- user inputs
+  that were never set, or an error from its last run -- rather than by eval
+  status, and the modal offers a one-off evaluation so that a block which has
+  never run can report either way. The `generate_code` plugin server takes
+  `update` in place of `visibility`, which is breaking for a front-end that
+  supplies its own `generate_code_server()` (#320).
 * `apply_board_update()` is now a real reducer rather than a no-op: its
   default `.board` method applies the core delta (block, link and stack
   mutations) to the supplied board and returns it, and update validation
@@ -89,7 +103,7 @@
   unfreezing resumes normal input handling (#231).
 * `background_construction_delay` now accepts `Inf`, skipping the background
   construction pass so a block is built only once it becomes required. Code
-  export ("Show code") then marks every block required, so the exported
+  export ("Show code") then claims every block on the board, so the exported
   script covers the whole board; an off-screen block that is not fully
   configured holds the export back instead of emitting broken code (#269).
 * Code export gates on the set of blocks that actually carry an expression,
--- a/R/board-server.R
+++ b/R/board-server.R
@@ -445,7 +445,7 @@ board_server.board <- function(id, x, plugins = board_plugins(x),
 
       call_plugin_server(
         "generate_code",
-        server_args = c(read_plugin_args, list(visibility = vis)),
+        server_args = edit_plugin_args,
         plugins = plugins
       )
 
--- a/R/plugin-code.R
+++ b/R/plugin-code.R
@@ -7,6 +7,18 @@
 #' by offering a download button, by providing this functionality as a
 #' `generate_code` module.
 #'
+#' Opening the modal asks for every block on the board to be built, through the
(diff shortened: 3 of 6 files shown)


nbenn:
Notes for whoever adapts this plugin under #321, since a few things were measured here that are not obvious from the code.

**The export needs blocks constructed, not evaluated.** It reads each block's `expr` and none of their results, and `lang` (returned as `expr`) is `reactive(exprs_to_lang(exp$expr()))`, not gated on `state_ready()`. A board whose merge block has no `by` still exports a complete script. This PR uses `sustain` only because it predates the `construct` component from #333; `construct` is what it actually wants, and it would leave the board lazy instead of putting the whole board in the eval set for as long as the modal is open.

**The configuration check does not need the eval set either.** With `construct` alone and an empty eval set, `board$blocks[[id]]$server$state_ready` reads `FALSE` for an unconfigured block and `TRUE` for a configured one. That is the check worth keeping: an unconfigured merge block exports `merge(a, b, by = character(0), ...)`, which does not error when run — it cross joins, 6 x 6 into 36 rows — so the harm is a silently wrong result rather than a crash, and `state_ready` catches exactly that case.

**What construction alone cannot give is the runtime-error check.** Reading `failed()` calls `res()`, so detecting a block that errors when it runs genuinely requires evaluation. Pairing `construct` with a one-off `evaluate` does not recover it: the drain observer's own status read is what pulls the evaluation, so the request completes and releases inside the same flush cascade, and by the time any output renders the statuses have reverted to `dormant`. Measured on a board with an unconfigured block, `construct` and `construct` plus `evaluate` are indistinguishable — empty eval set, `dormant`, gate reporting `ready` in both.

**One thing to know about `code_export_state()`.** It accepts `ready`, `dormant` and `stale`, but `block_eval_status()` returns the latter two only in its `!block_needed()` branch, so they are unreachable while anything holds the blocks. The tolerance therefore only fires when nothing has been requested, which is precisely when the verdict is meaningless. Tightening it to `"ready"` fails exactly one test in the suite, the one that documents the tolerance itself.

**A knock-on for the UI.** The Close button here exists so the server sees every close and can release on it. Nothing needs releasing after a `construct` request, so that constraint disappears and the modal can go back to `easyClose = TRUE` with no footer.

PR #324, opened 2026-08-14
Stop bbquote deleting a call's NULL elements

### LINKED ISSUE #323: bbquote() aborts on any expression containing a NULL call element
## Symptom

```
Error: 'names' attribute [4] must be the same length as the vector [3]
```

or, when the `NULL` slot is named:

```
Error: subscript out of bounds
```

This took down a **deployed** board on a view switch (blockr.viz's picker block,
whose expression defines two functions). It reproduces on `main` (`c7d9274`,
0.1.4).

## Cause

`R/utils-expr.R:84`, in `process_splices()` inside `bbquote()`:

```r
e_list[[i]] <- process_splices(e_list[[i]])
```

When `process_splices()` returns `NULL`, `[[<-` **deletes** the element rather
than replacing it. The call then comes back shorter than `names_e`, and the
following `names(e_list) <- names_e` aborts. With a named `NULL` the deletion
also shifts the remaining indices, so the loop runs off the end and gets
`subscript out of bounds` instead.

A `NULL` element in a call is legitimate and arises two ways:

**1. A literal `NULL` argument.** Fails everywhere, dev included:

```r
bbquote(f(x, NULL), list())
#> Error: 'names' attribute [3] must be the same length as the vector [2]

bbquote(list(a = NULL, b = .(x)), list(x = 1))
#> Error: subscript out of bounds
```

`f(data, NULL)` is an ordinary shape in a block expression.

**2. The srcref slot of a `function` definition.** This is the production-only
half. A `function` call has four elements and the fourth is its srcref, which is
`NULL` whenever the code was parsed *without* srcrefs — which is what an
**installed package** does (`keep.source = FALSE`) and what `load_all()` does
not:

```r
fn <- parse(text = "local({ f <- function(v, w) v; f(.(x), 1) })",
            keep.source = FALSE)[[1]]
bbquote(fn, list(x = 1))
#> Error: 'names' attribute [4] must be the same length as the vector [3]

# same text with keep.source = TRUE: works fine
```

So bquoting any expression that defines a function passed in every dev session
and crashed once installed.

## Fix

Single-bracket assignment of a one-element list replaces instead of deleting:

```r
e_list[i] <- list(process_splices(e_list[[i]]))
```

Branch `fix/bbquote-null-call-elements` (one commit, based on `f829543`, merges
cleanly into `main`). Verified: identical output to the current implementation on
every existing call shape (`.()`, `..()`, named splice, plain calls), all three
failing shapes above now pass, `test-utils-expr.R` green.

christophsax:
Fixes #323.

`e_list[[i]] <- NULL` deletes the element instead of replacing it, so any call
containing a `NULL` element comes back shorter than its names and `bbquote()`
aborts. Single-bracket assignment of a one-element list replaces rather than
deletes.

Two triggers:

- **A literal `NULL` argument** — `f(data, NULL)` is an ordinary shape in a
  block expression. Fails everywhere, dev included. A *named* `NULL` fails
  differently again (`subscript out of bounds`), because the deletion shifts
  the indices the loop is still walking.
- **A function definition's srcref slot.** A `function` call has four elements
  and the fourth is its srcref, which is `NULL` whenever the code was parsed
  without srcrefs — what an installed package does (`keep.source = FALSE`) and
  what `load_all()` does not. So bquoting any expression that defines a
  function passed in every dev session and crashed only once installed.

**Blast radius:** anything that bquotes a block expression defining a function
was broken in production only. The known casualty is blockr.viz's picker block
(two function definitions), which took down a deployed board on a view switch.

Verified identical output to the current implementation on every existing call
shape (`.()`, `..()`, named splice, plain calls); the three failing shapes are
covered by new tests in `test-utils-expr.R`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: blockr.core
 Title: Graphical Web-Framework for Data Manipulation and Visualization
-Version: 0.1.4
+Version: 0.1.5
 Authors@R: 
     c(person(given = "Nicolas",
              family = "Bennett",
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,13 @@
+# blockr.core 0.1.5
+
+* `bbquote()` no longer drops `NULL` elements while walking a call. Assigning
+  the result of the recursive step with `[[<-` deleted the element whenever it
+  was `NULL`, leaving the call shorter than its names and aborting. This hit any
+  expression containing a literal `NULL` argument, and — because a `function`
+  definition's srcref slot is `NULL` under `keep.source = FALSE` — any
+  expression defining a function, but only once the package was installed
+  (#323).
+
 # blockr.core 0.1.4
 
 * `apply_board_update()` is now a real reducer rather than a no-op: its
--- a/R/utils-expr.R
+++ b/R/utils-expr.R
@@ -81,7 +81,16 @@ bbquote <- function(expr, where = parent.frame(), splice = FALSE) {
       if (nchar(names_e[i]) && is_dots(e_list[[i]])) {
         names_e[i] <- ""
       } else {
-        e_list[[i]] <- process_splices(e_list[[i]])
+        # `e_list[[i]] <- NULL` DELETES the element; single-bracket
+        # assignment of a one-element list replaces it. A call element is
+        # legitimately NULL: the srcref slot of a `function` definition is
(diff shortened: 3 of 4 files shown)


christophsax:
Closing — not proposing this as a PR. The branch `fix/bbquote-null-call-elements` stays pushed for reference; see #323 for the diagnosis.

Issue #311, opened 2026-07-24
Extract a pure apply_board_update() shared by update validation and the reactive apply


nbenn:
## Motivation

blockr.core has no pure "apply an update delta to a board" function. The reactive apply (`apply_core_board_update()`) mutates `rv$board` imperatively, interleaved with UI work (`insert_block_ui()`, `construct_blocks()`, `session$destroy()`); core's `apply_board_update.board()` is a deliberate no-op, and the `apply_board_update()` generic serves only as the *extension* reducer (blockr.dock's `apply_board_update.dock_board()` applies views on top of the already-core-applied board).

Update *validation* would ideally ask "does applying this update yield a valid board?" by applying it to a scratch board and validating the result. #309 tried exactly that and it broke the blockr.dock revdep: on a `dock_board`, the `board_*<-` setters fire `validate_board.dock_board()`, which enforces view membership, but a core-level scratch apply never cascades views (dock does that in `apply_board_update.dock_board()`), so a valid block deletion got rejected. #309 therefore shipped the narrow fix — core validates only its own references via `validate_board_blocks_links` / `validate_board_blocks_stacks`, and dock validates views through its own `validate_board_update.dock_board()` + `NextMethod()`. That is the correct seam, but validation-by-applying stays out of reach until a pure, extensible apply exists.

## The change

Introduce a real `apply_board_update.board(board, upd) -> board` that applies the core delta (add blocks, fold link `mod` into `add`+`rm`, `modify_board_links`, `modify_board_stacks`, remove blocks last — the sequence `apply_core_board_update()` and #309's scratch apply both encode). Then route everything through it:

- `apply_board_update.dock_board()` calls `NextMethod()` for the core part, then applies views, so one generic yields the full board for any subclass. This is a **blockr.dock change** (cross-repo).
- `apply_core_board_update()` mutates the board object through the reducer and keeps only its UI side effects. This is the delicate part: those side effects are interleaved *between* today's data mutations (`construct_blocks()` needs the board already updated; teardown runs around removal), so the reactive flow has to be reordered around a single reduce.
- `validate_board_update_result()` becomes `validate_board(apply_board_update(augment_board_update(payload, board), board))` — apply the full core+extension update to a scratch board and validate it, composing across the seam. The narrow membership check then retires.

## Constraint

[...]

Issue #177, opened 2026-05-22
Stage-time link validator rejects payloads that `preprocess_board_update` would otherwise accept (dangling-link auto-cleanup mismatch)


nbenn:
## Repro

```r
library(blockr.core)

brd <- new_board(
  blocks = c(
    my_data = new_dataset_block("iris"),
    preview = new_head_block()
  ),
  links = c(lnk1 = new_link("my_data", "preview", "data"))
)

payload <- list(
  blocks = list(
    add = c(preview_new = new_head_block()),
    mod = list(),
    rm  = "preview"
  ),
  links = list(
    add = c(lnk_new = new_link("my_data", "preview_new", "data")),
    mod = list(),
    rm  = character()
  ),
  stacks = list(add = stacks(), mod = list(), rm = character())
)

validate_board_update(payload, brd)
# Error: Expecting all links to refer to known block IDs.
```

## Why

`validate_board_update_xrefs` builds the merged link set as
`board_links - links$rm + links$add` and the merged block set via
`combine_update_blocks(payload, board) = board_blocks - blocks$rm + blocks$add`.
With the payload above:

- merged blocks: `{my_data, preview_new}` (`preview` removed)
- merged links: `{lnk1: my_data → preview, lnk_new: my_data → preview_new}` (`lnk1` is in board, the payload didn't ask to remove it)

`validate_board_blocks_links` then rejects because `lnk1.to == "preview"` is not in the merged blocks.

But if you bypass the validator and just `update(payload)`, `preprocess_board_update` would extend `payload$links$rm` with the orphans of removed blocks (this is the existing auto-cleanup behaviour) and apply cleanly.

## Why it matters

This is the architectural shape of "replace a block": remove the old, add a new one in its place, wire a new link to the replacement. Today the consumer has to enumerate all incoming/outgoing links of the removed block and stage their removal too, duplicating logic that preprocess already does at apply time.

It bites the assistant's modify flow specifically: when a non-ctrl-able arg has to change, the recovery path is `remove_block + add_block(id_or_new)`, and any link that fed into the removed block needs explicit `remove_link` calls or the next `add_link` rejects. The LLM ends up tripping into this trap; see the trace at https://github.com/BristolMyersSquibb/blockr.assistant/pull/11 (look for the "Expecting all links to refer to known block IDs" hops).

## Suggested fix

Either:

1. Have the xref validator run against the preprocess-augmented payload (cleanest — single source of truth for "what will get removed").
[...]

PR #328, opened 2026-08-14
Apply a link mod in place instead of folding it into add + rm

### LINKED ISSUE #316: Modify a link in place instead of folding the edit into add + rm
## Motivation

A link edit — the `links$mod` delta the edit-link actions (dag / dock) emit — is not applied in place. `merge_link_mods()` resolves the delta to a fresh link object via `update_link()`, and both the reducer (`apply_board_update.board()`) and the reactive path (`apply_core_board_update()`) fold it into `add` + `rm`. So editing a link tears down and rebuilds the board row *and* the reactive wiring, even when only its `from` changed (the common "retarget" edit).

Where the reactive teardown costs:

- **Fixed-input `to` block** — light. `update_block_links()` runs `destroy_link()` then `setup_link()`: two assignments to one `rv$sources[[to]]` slot, with a transient `NULL` reaching downstream before the new value lands.
- **Variadic `to` block** — the real cost. `update_block_links()` calls `sync_dot_args()`, which drops *every* one of the block's `...args` input reactives and recreates them all, re-firing the whole downstream — even for a single link's `from` change. That is exactly the spurious re-evaluation the rest of core works to avoid.

## The change

Add an in-place `modify_link` path that updates the specific `rv$sources[[to]]` slot for the common edits and falls back to add/rm for structural changes:

- `from` change (retarget): a single `rv$sources[[to]][[slot]] <- new_from` re-fires just that input's `upstream_result`, nothing else.
- `input` change: a slot move (clear the old key, set the new).
- `to` change: a re-parent — keep as add/rm.

## Constraints

- **Variadic ordering.** A variadic block's dot-arg order is derived from its link set, which is part of why `sync_dot_args()` rebuilds wholesale. A targeted variadic update still has to reconcile arg order without a full resync, so the clean win is fixed-input retargets first; variadic needs more care.
- **The data side is trivial** (`board_links(board)[name] <- new_link`). That is what the `modify_board_links(mod=)` argument did; #311 retired it as unused and vctrs-unsafe on a `mod`+`rm` overlap, not because in-place is wrong. A real path wants a clean `modify_link` verb anyway.
- **Not profiled.** The motivation is the `sync_dot_args()` blast radius on variadic blocks, not a measured number — worth confirming with a trace before committing to the variadic case.

Follow-up to #311, which extracted the pure apply reducer but deliberately preserved the existing mod-to-add+rm fold.

nbenn:
## Summary

- A `links$mod` delta was resolved to a fresh link and folded into an add plus a removal, so editing a link tore the target block's input wiring down and rebuilt it. Where the edit keeps the slot the link already occupies — `to` and `input` unchanged — the existing slot is now pointed at the new source instead. An edit that moves the link (a changed `to` or `input`) still rewires through add + rm.
- The variadic case needs no arg-order reconciliation after all, so it is in scope rather than deferred. A fixed input's `upstream_result()` reads `src_rv[[input]]` and a variadic one reads `src_rv[[link_id]]`, while the dot-arg name comes from `input` — none of those keys move when only `from` changes, so the link set and its order are untouched.
- The `modify_board_links()` function regains the `mod` argument #311 retired, mirroring `modify_board_stacks()`, so the data side replaces a link in place rather than reconstructing its position through an add + rm overlap. An ID named in both `mod` and `rm` is replaced rather than removed — the rule `add` + `rm` already followed, and what keeps the #177 block-removal guard firing.
- The issue also asked for an `input` change to become a targeted slot move, and it did not get its own path. On a fixed-input target the rewire already *is* that slot move: `destroy_link()` clears the old key, `setup_link()` writes the new one, and no dot-arg resync runs on a non-variadic block. Only a variadic target still resyncs wholesale on an `input` rename, since renaming a dot arg is the arg-order reconciliation the issue flagged as needing more care — worth its own change if it proves to matter.

## What this actually costs, measured

The issue was filed on the premise of a `sync_dot_args()` blast radius, unprofiled. Measured against a pre-fix build (`sync` = `sync_dot_args()` calls, `keys` = invalidations of the dot-arg name derivation the block's expression reads, `eval` = re-evaluations of the target block), variadic target with two incoming links:

| payload | before | after |
| --- | --- | --- |
| `mod` changing `from` (retarget) | sync 1, wiring replaced, keys 1, eval 1 | sync 0, wiring kept, keys 0, **eval 1** |
| `mod` leaving `from` alone | sync 1, wiring replaced, keys 1, eval 1 | sync 0, wiring kept, keys 0, **eval 0** |
| `mod` renaming a dot arg (`input`) | sync 1, wiring replaced, keys 1, eval 1 | unchanged — falls back to add + rm |

So this is not a uniform win, and it is worth saying plainly which parts are which. Shiny coalesces invalidations within a flush, so a genuine retarget re-evaluates the target block exactly once before and after — the fold never multiplied downstream evaluation, and an earlier draft of this description claiming otherwise was wrong. What the fold did add is the rebuild itself (every `...args` reactive discarded and recreated, the block's arg-name derivation invalidated), and one wholly spurious re-evaluation whenever a mod leaves the wiring untouched. That last row is the only place an evaluation disappears, and it applies to fixed-input targets too, where the rest of the change is a no-op beyond dropping one redundant `reactiveValues` write. The data-side `mod` argument is a pure refactor — byte-identical results to the add + rm overlap it replaces.

Both effects are pinned by tests that fail against the pre-fix sources: one asserts the dot-arg reactives survive a retarget as the *same objects*, the other that a no-change mod costs zero evaluations and a real retarget costs one.

Fixes #316

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,16 @@
 # blockr.core 0.1.4
 
+* A link edit arriving as a `links$mod` board update is now applied in place
+  rather than folded into an add plus a removal. Where the edit keeps the
+  link's input slot -- a retarget that changes only `from` -- the target
+  block's existing input reactive is pointed at the new source, so only that
+  input re-fires. The fold previously tore the wiring down and rebuilt it,
+  which on a variadic target meant dropping and recreating every one of its
+  `...args` reactives, re-firing all of the block's data inputs for a single
+  link's edit. An edit that moves the link to another slot (a changed `to` or
+  `input`) still rewires. Correspondingly, `modify_board_links()` gains a
+  `mod` argument that replaces links in place, keeping each one's position
+  (#316).
 * `apply_board_update()` is now a real reducer rather than a no-op: its
   default `.board` method applies the core delta (block, link and stack
   mutations) to the supplied board and returns it, and update validation
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,15 +2,18 @@
 
 * A link edit arriving as a `links$mod` board update is now applied in place
   rather than folded into an add plus a removal. Where the edit keeps the
-  link's input slot -- a retarget that changes only `from` -- the target
-  block's existing input reactive is pointed at the new source, so only that
-  input re-fires. The fold previously tore the wiring down and rebuilt it,
-  which on a variadic target meant dropping and recreating every one of its
-  `...args` reactives, re-firing all of the block's data inputs for a single
-  link's edit. An edit that moves the link to another slot (a changed `to` or
-  `input`) still rewires. Correspondingly, `modify_board_links()` gains a
-  `mod` argument that replaces links in place, keeping each one's position
-  (#316).
+  link's input slot -- `to` and `input` unchanged -- the existing slot is
+  pointed at the new source, so the target block's input wiring survives the
+  edit. The fold previously tore that wiring down and rebuilt it, which on a
+  variadic target meant `sync_dot_args()` discarding and recreating every one
+  of the block's `...args` reactives and invalidating its dot-arg name
+  derivation. A genuine retarget re-evaluates the target block once either
+  way, so that part is unchanged; what the fold added on top was the rebuild,
+  plus a re-evaluation for a mod that leaves the wiring untouched -- that one
+  is now free, on fixed and variadic targets alike. An edit that moves the
+  link (a changed `to` or `input`) still rewires. Correspondingly,
+  `modify_board_links()` gains a `mod` argument that replaces links in place,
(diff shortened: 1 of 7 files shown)


nbenn:
Closing unmerged. Measurement against a build of `main` showed the motivating cost is not there: a genuine retarget re-evaluates the target block and its downstream exactly once before and after this change, because shiny coalesces invalidations within a flush. The only payload that improves is a `mod` that changes nothing, which is a no-op by construction rather than a real edit.

Full numbers and the two corrections to the issue's premise are in #316, which is closed with them.

PR #329, opened 2026-08-14
Gate block control updates on front-end UI readiness

### LINKED ISSUE #317: Blocks on a non-startup view lose the restored state of every control fed by update*Input()
A block whose panel is not on the startup view still gets a server at boot,
because blockr.core constructs it whenever a visible block downstream needs it.
That server pushes its restored state to the client at module start — but the
panel's HTML does not exist yet, so `updateSelectInput()` and friends address an
unbound input and Shiny drops the message. Nothing re-sends it. The block
computes correctly and everything downstream is right; only its controls come up
blank, and they never recover.

`new_merge_block()` shows it with no third-party code involved. Reproduced on
blockr.core 0.1.4 + blockr.dock 0.1.3 (installed, no `load_all()`).

**Not urgent** — see "Priority" at the bottom. Filing it because the analysis is
fresh and because the fix is a small seam that other packages keep hand-rolling.

## Repro

```r
library(blockr.core)
library(blockr.dock)

control <- FALSE  # set TRUE for the working comparison

board <- new_dock_board(
  blocks = c(
    a   = new_dataset_block("iris"),
    b   = new_dataset_block("iris"),
    mrg = new_merge_block(by = "Species"),
    out = new_head_block(n = 5L)
  ),
  links = links(
    from  = c("a", "b", "mrg"),
    to    = c("mrg", "mrg", "out"),
    input = c("x", "y", "data")
  ),
  grids = if (control) {
    list(shown = dock_grid("a", "b", "out", "mrg"), hidden = dock_grid())
  } else {
    list(shown = dock_grid("a", "b", "out"), hidden = dock_grid("mrg"))
  },
  active = "shown"
)

serve(board)
```

Switch to the "Hidden" view.

* **Expected**: "By columns" shows `Species`.
* **Actual**: "By columns" is empty — while the merge block's own output right
  below it is joined on Species (`Species`, `Sepal.Length.x`, `Sepal.Length.y`).
  Probing the DOM: `select#<ns>-block_mrg-expr-by` has `opts: []`, `sel: []`.

`control <- TRUE` moves the same block to the startup view; the control then
reads `Species` (`opts: ["Species"]`). Same board, same state, only the view
membership differs.

## Mechanism

`new_merge_block()`'s `ui()` is `selectInput(inputId = NS(id, "by"), choices =
list(), multiple = TRUE)` — no `choices`, no `selected` — so 100% of its state
reaches the client through `observe(updateSelectInput(session, "by", choices =
cols(), selected = sels()))` in the block server (`R/transform-merge.R:66-73`).
On a deferred panel that one push is dropped, and neither `cols()` nor `sels()`
invalidates again, so it is never re-sent.

The core blocks that look healthy only survive by accident of drafting: they
bake the state into the static `ui()` (`new_dataset_block()` passes `selected =
dataset`, `new_subset_block()` passes `value = subset`), so the value is in the
HTML whenever the panel is finally inserted.

`new_scatter_block()` shows a second face of the same race: its
`observeEvent(cols(), updateSelectInput(...))` (`R/plot-scatter.R:43-59`) fires
once at boot, before dockview has inserted *any* panel, so its X/Y dropdowns
never grow past the single restored value — **including on the startup view**.
A block whose column picker cannot offer another column is worth fixing on its
own, independent of anything below.

## Why this belongs in core, not in the blocks

The same bug has been hit in blockr.dm (crossfilter, value filter, table
picker), blockr.dplyr (filter and friends) and blockr.io (path inputs), and each
one has been patched with a hand-rolled client-side replay queue plus a
`Shiny.setInputValue(el.id + '_ready', ...)` handshake so R can re-send. That
workaround needs JavaScript of our own to hang the announce on — so a block that
uses nothing but `updateSelectInput()`, which is `merge_block` and most of
blockr.dm's plain blocks, **cannot be fixed that way at the block level at all**.

Core also already owns the missing piece. `blockr.dock` reports which cards have
painted (`mark_cards_rendered()` → `visibility$visible[[id]](TRUE)`), core holds
it as `vis$visible[[id]]` (`R/board-server.R:645`), and core already uses it to
gate *output rendering* (`render_gate_observer()`, `R/block-server.R:603`). It
just never reaches the block's own server: `block_server.block()` calls
`expr_server(x, data)` and `expr_server.block()` calls the constructor's server
with `id = "expr"` plus the data arguments only (`R/block-server.R:155` and
`:443`). The visibility environment stops one level above the code that does the
restore pushes.

## Suggested fix

Pass a `block_ui_ready` (or `visible`) reactive down through `expr_server()`
into block servers that opt in, e.g. by declaring an extra formal. A block that
takes it wraps its restore pushes in `observeEvent(ui_ready(), ...)` instead of
pushing at module start. Small, opt-in, and it turns the handshakes we keep
hand-rolling into one supported seam. `merge_block` and `scatter_block` would be
the first two consumers.

Two alternatives considered and not preferred: core buffering and replaying a
block's outbound messages until the card is confirmed painted (no block changes,
but core has to intercept `session$sendCustomMessage` and `update*Input` for the
module — invasive); and fixing delivery in dockViewR, which inserts panel HTML
with `innerHTML` (`srcjs/modules/components.js:20`) so `Shiny.renderContent`
would make panel content bind on insertion — worth doing anyway, since it is
also why inline `<script>` in a block `ui()` never runs inside a dock panel, but
it does not by itself fix a push that happened before the panel existed.

## Workaround that works today, without JS and without a core change

Render the controls instead of pushing them: `ui()` is `uiOutput(NS(id,
"ctrl"))` and the server does `output$ctrl <- renderUI(selectInput(...,
selected = isolate(state())))`. Shiny suspends an output observer while its
element is absent from the DOM and resumes it once the element binds, so a
rendered control is deferred-safe for free.

Verified side by side: two merge blocks on the same deferred view, the stock one
blank, the `renderUI` one restored. This does not cover controls that are custom
widgets driven by `sendCustomMessage` — re-rendering those re-mounts the widget
and loses its client state — so those still need the seam above.

## Priority

Not urgent, for three reasons:

1. The `renderUI` workaround above needs no core change and no JS, and it covers
   exactly the case that is otherwise unfixable at the block level. Affected
   blocks can be fixed one at a time in their own packages.
2. No data is wrong. The pipeline computes correctly and every downstream block
   is right — the failure is confined to how a deferred block's own controls
   display.
3. In core itself only `merge_block` and `scatter_block` are exposed; the rest
   bake their state into the static `ui()` and are unaffected.

The one piece that is arguably worth doing sooner is the `scatter_block` half,
since it misbehaves on the startup view too, and it is a small local fix rather
than the seam.

Screenshots of both states and a fuller cross-package write-up exist locally —
happy to attach if useful.

nbenn:
## Summary

Block servers can declare a reserved `ui_ready` argument, which `expr_server()` fills with a reactive that is `TRUE` while the front-end reports the block on screen. It reads the `visible` channel that already drives the render gate, so no new channel is plumbed, and it is constantly `TRUE` when nothing gates visibility. Guarding a push with `req(ui_ready())` holds it until the controls exist and re-sends it whenever the block returns to screen.

Like `...args`, the argument is reserved rather than a data input: `block_inputs()`, `block_arity()` and the `dat_valid` arity check all skip it, so declaring it neither opens a link slot nor breaks the validator.

Adopted by `merge_block`, `scatter_block` and `head_block`. The issue names the first two; `head_block` turned out to carry the same defect in the row bound it pushes (`max = nrow(data())`), sent once at boot and dropped on a deferred panel the same way.

## What I measured

The seam itself is verified end to end against blockr.dock 0.1.3 and dockViewR 0.4.0. An instrumented off-screen block logs `ui_ready=FALSE` at boot, `ui_ready=TRUE` roughly 1.4 s after its view is opened, pushes there, and its select comes up populated.

The reported symptom, though, did **not** reproduce on that stack. Running the issue's verbatim repro against `main`, the merge block's "By columns" is blank for about 2 s after the switch to the Hidden view and then fills in on its own — it does not stay blank. An A/B on a single board (one block gated, an otherwise identical one not) shows why: the ungated block's push observer re-fires when its upstream re-publishes on the view switch, so the dropped boot-time push gets repaired by accident. The defect is directly observable either way — the ungated block logs its push at boot with no panel in the DOM, and its control is empty for as long as the panel is absent — but whether it stays broken depends on whether some upstream happens to invalidate later. The gated block does not depend on that luck.

The unit tests carry the regression teeth: each of the three block tests fails on the unguarded code at the "no push while hidden" assertion, and the board-level test fails if `ui_ready` is stubbed to a constant `TRUE`.

Fixes #317

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -191,6 +191,20 @@
   is `NULL` under `keep.source = FALSE` -- any expression defining a function,
   but only once the package was installed rather than loaded with `load_all()`
   (#323).
+* Block servers can opt into a UI-readiness handshake by declaring a `ui_ready`
+  argument, which `expr_server()` fills with a reactive that is `TRUE` while
+  the front-end reports the block on screen. Controls whose state reaches the
+  client through `update*Input()` were pushed at module start, but a front-end
+  that defers panels builds an off-screen block (because something downstream
+  needs its result) long before its UI exists, so shiny dropped the message
+  against an unbound input and nothing re-sent it: a block on a non-startup
+  view came up with blank controls while computing correctly. Guarding the
+  pushes with `req(ui_ready())` holds them until the controls exist and
+  re-sends them whenever the block returns to screen. Like `...args`, the
+  argument is reserved and counts towards neither `block_inputs()` nor
+  `block_arity()`. Adopted by `merge_block`, `scatter_block` and `head_block`
+  -- the last two also pushed their column choices and row bound before first
+  paint, so they misbehaved on the startup view too (#317).
 
 # blockr.core 0.1.3
 
--- a/R/block-class.R
+++ b/R/block-class.R
@@ -332,7 +332,7 @@ validate_block_ui <- function(ui) {
 
 validate_data_validator <- function(validator, server) {
 
-  server_args <- setdiff(names(formals(server)), "id")
+  server_args <- server_data_args(server)
 
   if (!length(server_args) && not_null(validator)) {
     blockr_abort(
@@ -549,7 +549,10 @@ c.block <- function(...) {
 #' inputs like for example a block providing [base::rbind()]-like
 #' functionality), `block_arity()` returns `NA` and the special block server
 #' function argument `...args`, signalling variadic behavior is stripped from
-#' `block_inputs()`.
+#' `block_inputs()`. A second reserved server function argument, `ui_ready`
+#' (see [expr_server()]), carries the front-end readiness of the block UI
+#' rather than data and is stripped from `block_inputs()` and `block_arity()`
+#' as well.
 #'
 #' @section External control:
 #' Blocks can expose constructor inputs for programmatic control from outside
@@ -706,7 +709,11 @@ block_inputs.block <- function(x) {
 }
 
 block_expr_inputs <- function(x) {
-  setdiff(names(formals(block_expr_server(x))), "id")
+  server_data_args(block_expr_server(x))
+}
+
+server_data_args <- function(server) {
+  setdiff(names(formals(server)), c("id", "ui_ready"))
 }
 
 block_ctor_inputs <- function(x) {
--- a/R/block-server.R
+++ b/R/block-server.R
@@ -92,6 +92,20 @@
 #' `gate_visibility` [blockr_option()] (default `TRUE`) turns gating off
 #' entirely.
 #'
+#' The `visible` channel also drives the point at which a block may talk to its
+#' own controls. A block server that ships state to the client with
+#' [shiny::updateSelectInput()] and friends cannot do so at module start: on a
+#' front-end that defers panels the block is built (because something needs its
+#' result) long before its UI exists, the update addresses an unbound input and
+#' shiny drops the message, and nothing re-sends it. A block opts into the
+#' handshake by declaring a `ui_ready` argument on its server function --
+#' reserved, like `...args`, so it is not mistaken for a data input --
+#' whereupon [expr_server()] hands it a reactive that is `TRUE` while the
+#' front-end reports the block on screen (and constantly `TRUE` when nothing
+#' gates visibility). Guarding the pushes with `req(ui_ready())` holds them
+#' until the controls exist and re-sends them whenever the block returns to
+#' screen. Core's `merge_block`, `scatter_block` and `head_block` use it.
+#'
 #' The same bundle carries a third channel, `frozen`, through which a
 #' front-end reports the blocks whose inputs it has hidden (for example a
 #' locked board that shows outputs but not controls). While frozen a block is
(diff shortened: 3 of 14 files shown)


nbenn:
Superseded by the design recorded in #317, closing unmerged.

This branch added an opt-in `ui_ready` reactive so a block could hold its `update*Input()` pushes until the front-end reported it on screen. It works, and the seam is verified end to end against blockr.dock 0.1.3 — an instrumented off-screen block logs `ui_ready=FALSE` at boot, `TRUE` about 1.4 s after its view is opened, pushes there, and its control comes up populated.

The reason to drop it anyway is that it keeps paint on the event channel. Every block has to opt in, forever, to go on using `update*Input()` for something that is not a transition. Two mechanisms cover the same ground without that: painting static block UI at construction (BristolMyersSquibb/blockr.dock#438) fixes every static-UI block including third-party ones with no code change, and a server-supplied UI (#346) is lazy and correct by construction for controls whose content depends on input data.

Also worth recording from this branch: `merge_block()` discards its `by` argument in its UI function, which is a bug independent of paint timing and is now #345.

Issue #333, opened 2026-08-16
Construction demand has no home in the board update payload


nbenn:
The board update payload can ask for a block to be evaluated, through `evaluate` or `require`, but not for one to be merely built. Construction follows evaluation as a side effect, so a consumer that needs a block *present* has to make it *run* as well.

The two inputs to construction are visible in `needed_block_ids()`:

```r
union(need, ever_required(required))
```

The first is the eval set, which is where `evaluate` and `require` land. The second reads a non-`NA` `vis$required` slot — the "build this, but nothing needs it evaluated" state that #321 identifies as job (2) of that tri-state. Core writes `vis$required` in exactly one place, `add_vis_slots()` setting `NA`; every non-`NA` write comes from a front-end or a callback. So "build it, don't run it" is a capability the front-end has and nobody else does.

## What it costs

The code export is the worked example. It needs every block **constructed**, because `export_wrapped_code()` reads each block's `expr` and an unconstructed block contributes none — that is the `` `NA` <- `` junk #300 fixed. It does not need any block **evaluated**: `lang` (exported as `expr`) is `reactive(exprs_to_lang(exp$expr()))`, not gated on `state_ready()`, and a board whose merge block has no `by` still exports a complete script.

With no way to say "construct", #322 claims `require` for the modal's lifetime instead, so opening "Show code" evaluates the whole board. That is a milder version of the behaviour #320 exists to fix, bounded by how long the modal stays open. A construction request would let the export hold what it actually needs and leave the board lazy, with evaluation happening only if the user asks for it.

## Suggested shape

```r
update(list(construct = board_block_ids(board$board)))
```

A bare character vector, like `evaluate`. No owner, no release, no state retained: once a block is built it stays built, so there is nothing to hold and nothing to hand back. The three request components then differ in exactly one axis:

| component | who lets go |
| --- | --- |
| `evaluate` | core, once the block has run |
| `require` | the owner named in the payload |
| `construct` | nobody — there is nothing to release |

Semantics:

- Joining `update_request_components()` gives it the lock behaviour of the other two: it carries no state change, so a locked board still accepts it.
- Validation matches `evaluate` — a character vector whose ids resolve against the post-update block set, so one payload can add a block and construct it. The check is near-identical to `validate_board_update_evaluate()` and wants to be one shared helper rather than two.
- Requesting a block that is already built is a no-op.
[...]

PR #337, opened 2026-08-18
Fold front-end demand into the one multi-owner claim set

### LINKED ISSUE #321: Evaluation demand should be one multi-owner set, not two channels

nbenn:
## Summary

Evaluation demand was expressed two ways depending on who asked: a front-end wrote the per-block `visibility$required` tri-state, while everyone else went through the `sustain` / `evaluate` / `construct` payload components. This retires the front-end's channel — it now claims under an owner label like any other consumer, so core no longer distinguishes its demand from a code export's and the demand axis cannot silently become multi-writer the way `required` did (the defect behind #320).

The tri-state was doing three jobs. Only evaluation demand folds into the claim set; the other two are separated rather than absorbed:

- **Gating activation** becomes an explicit declaration — a front-end writes its owner label into the new board-wide `visibility$gate` channel. Inferring it from claims instead would mean a consumer asking about one block flips an ungated board into gated mode and parks every other block, which `test-visibility-gating.R` now pins.
- **Construction demand** is the `construct` component from #333.
- **Paint acknowledgement** on `visible` is untouched, so the demand-versus-paint comparison that keeps background construction from competing with first paint stays meaningful. It is now scoped to the gating owner's own claim, since a claim held by anyone else names blocks nobody is putting on screen and would stall the backlog for the session.

The gate declaration is written synchronously as the callback is set up rather than travelling in a payload, and that is forced rather than chosen: a payload applies at `priority = -Inf`, after the first flush has already decided what to construct. Measured on this branch with demand arriving only by payload, flush 1 built every block on the board and left the needed set at `TRUE`. Demand itself has no such constraint, which is why it rides `sustain`.

One consequence worth flagging: a front-end that has declared the gate but whose opening claim is still in flight is indistinguishable from one that claims nothing, so the background pass now waits for that first claim to land (`rv$gate_claimed`). Production previously satisfied this only by accident of the pacing delay; making it explicit is what keeps the priority lane ordered when the claim arrives by payload.

## Migration

| Before | After |
| --- | --- |
| `visibility$required[[id]](TRUE)` | `update(list(sustain = list(<owner> = list(add = id))))` |
| `visibility$required[[id]](FALSE)` | `update(list(sustain = list(<owner> = list(rm = id))))`, plus `update(list(construct = id))` to keep it built |
| `visibility$required[[id]]` non-`NA` activates gating | `visibility$gate("<owner>")`, once, as the callback is set up |
| `visibility$visible[[id]](…)`, `visibility$frozen[[id]](…)` | unchanged |

Breaking for blockr.dock, which drives `required` in `mark_cards_built()` and `show_cards()`; blockr.dock#417 tracks the adoption. The two neighbouring unreleased 0.1.4 NEWS entries (#333, #320) described the change in terms of the `required` channel and were corrected in place.

Fixes #321

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,19 +1,34 @@
 # blockr.core 0.1.4
 
+* Evaluation demand is now one multi-owner set rather than two channels. The
+  front-end's per-block `required` channel is gone: what it needs evaluated is
+  a `sustain` claim held under an owner label like any other consumer's, and
+  what it needs merely built is a `construct` request. Core no longer
+  distinguishes a front-end's demand from a code export's, and because no owner
+  writes another's claim, the channel cannot silently become multi-writer the
+  way `required` did. The three jobs the tri-state used to do in one slot are
+  now separate: gating activation is an explicit declaration a front-end makes
+  by writing its owner label into the new board-wide `visibility$gate` channel
+  -- inferring it from claims instead would let a consumer asking about one
+  block park every other block on an ungated board -- and construction demand is
+  the `construct` component. Reporting paint on `visible` is unchanged, and the
+  background pass still holds until every block the gating front-end claims is
+  reported painted. Breaking for any front-end that drives
+  `visibility$required` (#321).
 * Core's own board UI now drives visibility, through a board callback like any
   other front-end rather than from inside the board server. Stacks render as an
   accordion which opens one stack and collapses the rest, so on a stacked board
   part of what is on screen was hidden from the first render while every block
   evaluated and rendered regardless -- nothing read the input `bslib` had
   already wired for reporting which stacks are open. The new `gate_stacks()`
-  callback marks the blocks of every open stack plus every unstacked block
-  required and parks the rest, so collapsing a stack stops its blocks
-  evaluating and expanding one starts them again. Parked rather than dropped: a
-  collapsed stack's blocks stay built, so re-expanding shows them without a
-  rebuild. It is `board_server()`'s default `callbacks` value and gates nothing
-  until that accordion reports, so a board driven by another front-end -- which
-  passes its own callbacks, and whose UI never binds the input -- is left
-  alone; a consumer that wants both keeps it in the list,
+  callback claims the blocks of every open stack plus every unstacked block and
+  parks the rest, so collapsing a stack stops its blocks evaluating and
+  expanding one starts them again. Parked rather than dropped: a collapsed
+  stack's blocks stay built through a `construct` request, so re-expanding shows
+  them without a rebuild. It is `board_server()`'s default `callbacks` value and
+  gates nothing until that accordion reports, so a board driven by another
+  front-end -- which passes its own callbacks, and whose UI never binds the
+  input -- is left alone; a consumer that wants both keeps it in the list,
   `callbacks = list(gate_stacks(), my_callback)`. The `gate_visibility` option
   turns it off along with all other gating. The accordion container ID moves
   from `<board>_stacks` to the board-namespaced `<board>-stacks`, which is what
@@ -26,11 +41,10 @@
   long as it needed the expressions. Unlike `evaluate` and `sustain` it retains
   no state: a built block stays built, so there is no owner to name and nothing
   to release, and requesting a block that is already built does nothing. The
-  request joins neither the eval set nor the front-end's `required` channel, so
-  it cannot turn a lazily evaluating board into an eagerly evaluating one
-  (#333).
-* Showing the generated code no longer writes the front-end's `required`
-  channel. A block parked with `required[[id]](FALSE)` was overwritten and
+  request joins neither the eval set nor the gate declaration, so it cannot
+  turn a lazily evaluating board into an eagerly evaluating one (#333).
+* Showing the generated code no longer writes the front-end's own demand
+  channel. A block the front-end had parked was overwritten and
   never restored, so a single "Show code" turned a lazily evaluating board
   into an eagerly evaluating one for the rest of the session, with nothing
   left to release it. The export asks for construction alone through the
--- a/R/block-server.R
+++ b/R/block-server.R
@@ -58,39 +58,49 @@
 #' [block_output()] generic. The [block_ui()] generic can then be used to
 #' control rendering of outputs.
 #'
-#' A front-end (such as blockr.dock) drives per-block channels that
-#' [board_server()] hands to the board callback as `visibility`. Two of them
-#' gate what is built and shown: `required` (which blocks it needs built and
-#' evaluated) and `visible` (which blocks it has arranged on screen).
-#' Requirements are a cause the front-end -- and
-#' core-side features such as code export -- declare; visibility is the effect
-#' the front-end reports back once it has painted a block. Rendering is gated
-#' on `visible`: the render observer is suspended while a block carries no
-#' visible slot and resumed once the front-end writes a non-empty string for
-#' it, starting suspended so nothing renders before the first report.
-#' Evaluation is gated on the *needed* set, the `required` blocks together with
-#' their upstream closure over [board_links()] (recomputed only when
-#' requirements or links change). A block's input data reactives stay
-#' unfulfilled (they [shiny::req()] out) unless the block is needed, so a block
-#' that is neither required nor feeding a required block pulls no input and
-#' stays fully quiescent: its result reactive, and any observer its expression
-#' server registers on the incoming data, all short-circuit and do nothing. A
-#' needed but off-screen block (one feeding a required block) evaluates but
-#' does not render. Block-server *construction* is prioritized the same way:
-#' the needed set is instantiated first so that first paint waits only for the
-#' required blocks and their upstreams, and the remaining block servers are
-#' built progressively in the background. That background pass holds until the
-#' front-end reports every required block as visible, so it never competes with
-#' first paint. A `required` slot of `FALSE` keeps a block built but dormant
-#' (ever required, not needed now); an absent slot leaves it unbuilt. Until a
+#' A front-end (such as blockr.dock) declares that it will drive visibility by
+#' writing an owner label into the `gate` channel of the `visibility` bundle
+#' that [board_server()] hands to the board callback. That declaration, and
+#' nothing else, is what flips the board from evaluating everything to
+#' evaluating only what is needed; with no front-end every block is needed and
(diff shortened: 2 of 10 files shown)


nbenn:
Core's own front-end now drives the visibility channels through a board callback rather than from inside `board_server()`, which changes what this PR has to carry. The stack accordion is `gate_stacks()`, and it is `board_server()`'s default `callbacks` value; a board driven by another front-end passes its own callbacks, so core tracks nothing and never competes for the same slots. Dock already does exactly that from `blockr_app_server.dock_board()`, so nothing is needed on that side. Landed on `main` in #339.

Worth carrying into the vocabulary here: whatever replaces `required`, *who tracks* is then answered by which callbacks the board was given, not by a runtime check inside `board_server()`. That makes it static, so there is no declaration ordering to get right, and a front-end cannot end up sharing a channel with core by accident.

That callback is a second migration site alongside blockr.dock#417, and a small one — its whole body is one `required` write and one `visible` write, and both collapse into a `gate` declaration plus a `sustain` claim under the same label. It could not be written that way against `main`: measured there, a callback that sends only a `sustain` claim leaves `gating_active()` at `FALSE` and every block needed, because activation reads `has_required()` alone, and the `construct` test in `test-visibility-gating.R` pins that a request must not flip an ungated board. Worth a line in the migration table so the core-side consumer is not missed.

One asymmetry to flag, because it runs the other way. The stack callback states what it requires only from an observer, never synchronously as it is set up, and that is fine against `required`: `required[[id]](FALSE)` still counts as ever-required, so a collapsed stack's blocks stay in the construction set either way, and nothing evaluates in the flush-1 transient because rendering is gated on `visible`, still `NA`. Under a claim set they are simply absent from the claim, so the first tick of `construct_blocks_in_background()` would take its ungated fast path, build the whole board and destroy the ticker. The synchronous declaration this PR already requires is what covers that, and a core-side owner will need it for the same reason a front-end does — worth saying explicitly, since "state it as you are set up" reads like a front-end-only obligation today.

nbenn:
On retiring `visible` as well: I probed an ungated core board on `main` with the second stack collapsed, and the result argues against that channel rather than for it.

- The eval status of `d`, in the collapsed stack, is `ready` — it evaluated.
- Shiny's own `session$clientData[["output_my_board-block_d-result_hidden"]]` is `TRUE`, so shiny knows the element is hidden.
- The rendered output is in the DOM anyway.

The reason is the ordering: the card is rendered into `#my_board_blocks` while visible, then moved into the collapsed accordion body by the `move-block-ui` message, so the hidden report arrives after the first render and `suspendWhenHidden` never gets its chance.

So neither shiny's flag nor a front-end's acknowledgement is early enough to stop the first render. The only signal that is early enough is what the front-end has already declared it intends to show — which is demand. Gating rendering on `visible` works because the acknowledgement carries the same information one round trip later, not because it says anything demand does not. As a render-gate input it looks redundant.

Two jobs I would not fold in with that, because I have not measured them:

- **The background-construction hold.** `required_fulfilled()` is `visible`'s only other consumer in core, and it is already weak — `schedule_construction()` arms through `onFlushed()` plus `later()`, so it waits for a genuine idle window regardless, and this PR's own notes record that production satisfied the claim latch only by accident of the pacing delay. My guess is it drops out. It is a guess.
- **The `NA` state.** Dock reads `ls(visibility$visible)` with `!is.na` as its build ledger (`built_cards()` / `slot_built()` in `R/block-ui.R`), which is dock state living in a core channel. Retiring `visible` means dock keeps its own ledger. That is a dock change rather than a core requirement, but it should be named, or the retirement reads cheaper than it is.

nbenn:
To state the end point the two comments above only circle: once `required` is gone here and `visible` goes the way argued above, the remaining reason to hand a callback the `visibility` bundle is `frozen`. Move that too and the bundle stops being part of the callback signature altogether — it becomes core-internal state written by the applied payloads, and `update` is the whole front-end interface. That seems worth stating as the target, because each channel argued on its own reads like three separate tidy-ups rather than one retirement.

The `frozen` channel has exactly one consumer in core (`block-server.R`, the input-freeze guard) and one writer in blockr.dock (`freeze_hidden_inputs()`, driven off the card's section state and the board lock). Unlike paint it is not redundant: only the front-end knows which controls it has hidden, so core cannot derive it. It moves rather than disappears — an owner-keyed `update` component of the same shape as `sustain`, since what it expresses is "this owner asserts these blocks are read-only".

One cost worth naming rather than discovering later: dock already folds two `update` writers into a single payload per flush, and routing `frozen` through the same channel makes a third. That is a dock-side change with its own edge cases, not a free relabelling.

The part that genuinely resists is the gate declaration, and it is the reason the bundle cannot simply be deleted. A payload applies at `priority = -Inf`, after the first flush has decided what to construct, which this PR measured — so the declaration cannot ride `update` the way demand can. My suggestion is to make it a property of the board rather than a runtime write: a `dock_board` states that an external front-end drives visibility, core reads it as it sets up, and there is no ordering left to get wrong. Core's own accordion callback then needs no declaration at all, since it is the default and the board says nothing. The alternative — keeping one narrow declaration handle in the callback signature while the three channels leave — works too, but it keeps a foot in the API this is trying to retire.

nbenn:
Heads up from #343 / #344: this branch reintroduces the load-time evaluation that #343 reports.

Measured on `321-unified-demand` as of `Port core's stack gating onto the claim set`, running this branch's own `inst/examples/board/gate/app.R` with a counter registered on `block_eval` for `dataset_block`:

```
evaluated at load : datasets::BOD datasets::ChickWeight
status_b          : ready
status_d          : dormant
```

The block in the collapsed stack (`c`, stack `s2`) runs once and then parks. The mechanism is the same three lines as on `main`: `gate_stacks()` writes `vis$gate(owner)` inside the observer, only once `stacks_reported()` is true, so through the first flush `gating_active()` is `FALSE` while `rv$needed` is still its initial `TRUE`, and everything built is evaluated.

That is a small fix on its own — the seed from #344 ports over — but it points at something this PR comes close to solving and stops just short of.

## The declaration is right; the timing is not

Replacing "has anyone written demand" with `not_null(vis$gate())` is the fix for the real ambiguity, and it is the part worth keeping: the presence of a front-end becomes something declared rather than inferred from the presence of its data.

What stays inferred is *when*. Board updates apply at `priority = -Inf`, after the flush's priority-0 observers, so a front-end whose only channel is `update` cannot declare in time by construction. The requirement to declare synchronously therefore remains an unwritten convention that every front-end has to know and that nothing enforces — and it is precisely what blocks retiring `visibility` in favour of the general-purpose update mechanism downstream (BristolMyersSquibb/blockr.dock#417).

The `rv$gate_claimed` latch is a second symptom of the same gap: it exists because an empty claim and no claim are indistinguishable, which is only a question because core does not know a claim is coming.

## Registration is the moment core can know

The callback list is the one thing `board_server()` holds synchronously, before any flush and without a round trip. Letting a callback declare itself the gate owner *there*, rather than by sending anything, closes the window without a convention:

- The needed set starts empty when a gate owner is registered and `TRUE` when none is, which is the question actually being asked.
- The `gate_claimed` latch goes, since an empty claim set then reads as "not yet" rather than "nobody".
- Core's own `gate_stacks()` needs no synchronous seed at all, and most of #344's first commit dissolves.
- A front-end can move to `update` alone, which is what BristolMyersSquibb/blockr.dock#417 wants.

Happy to take this on if you would rather it were folded in here than tracked separately.

